PR #8, opened 2014-03-04
Implemented Multi-Currency Support (or at least the start of it)

davzie:
All tests passing. I haven't been able to test directly with Stripe just yet though...

Signed-off-by: David Thorpe davzie@davzie.com


commandantp:
any updates on that?

commandantp:
Why are we not simply passing the currencies as arguments, it will be passed as an option to Stripe anyways ? That would allow to support any currency.

davzie:
Mate, this was so long ago I haven't a clue what's happening right now, I'm just drinking some honey and hot water like a boss. I think that Cashier is getting some TLC with the upcoming release of Laravel 5.2. I could be wrong though.

PR #92, opened 2014-07-21
Not fetching last four digits of a card if there's no default card

ashitvora:
Stripe allows us to create a customer without any card.
In that case trying to fetch last four digits of the default card after user is subscribed to a plan throws error.
This is possible when user is trying to subscribe to a free plan and app doesn't require them to enter their credit card info.


scotself:
Sorry, maybe I missed something...is this included in an update? It seems like it is in the current codebase, but want to be sure so I can stop excluding from composer.json on a specific project. Thanks if it is indeed included now.

PR #100, opened 2014-08-17
Get the entity's upcoming invoice

djug:


lostincode:
@taylorotwell are you able to tag a version for this so we don't have to use master?

PR #126, opened 2014-11-29
General charges on a customer (one time purchases)

davidxd33:
As a remake of #96, this will introduce general charges on a given card or customer aka one time purchases.

```php
$user = Auth::user();
$token = Input::get('stripeToken');

$user->card($token)
     ->charge([
         "amount" => 1500,
         "currency" => "usd",

         // additional properties...
     ]);
```

If you omit the token, it will try an search for a card stored on the customer object else throw an exception. When a charge is successful, it will return a charge object that you can catch and store. 

If a customer doesn't exist, it will create the customer with/without a card. If without a card, it will store the customer as well as throw an exception once a charge is made without a card. 

charge() takes an optional second parameter for defining user properties.

### DIFF
--- a/src/Laravel/Cashier/BillableTrait.php
+++ b/src/Laravel/Cashier/BillableTrait.php
@@ -17,7 +17,7 @@ trait BillableTrait {
 	/**
 	 * Manipulate a card or a card in storage.
 	 * 
-	 * @param  string|null $token
+	 * @param  string|null  $token
 	 * @return \Laravel\Cashier\StripeGateway
 	 */
 	public function card($token = null) {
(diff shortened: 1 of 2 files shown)


srtfisher:
+1

jasondavis:
+1 this or any other implementation that adds this simple feature finally!  Pleeeease somebody!

auro1:
+1 perhaps this could also be integrated as invoiceitems? so billing could still only happen once a month

kallepa:
+1

GrahamCampbell:
The merge conflicts need resolving.

smick:
+ 1

DraperStudio:
:+1:

BrianDillingham:
+1

davidxd33:
This will likely not be merged but will be implemented in another form in the future; possibly in Laravel 5.1+.

See https://github.com/laravel/cashier/pull/96#issuecomment-67557549

taylorotwell:
Done a little differently.

PR #134, opened 2015-01-24
Cashier 3.0 is L5 only

GrahamCampbell:


GrahamCampbell:
Also, dependencies with no upper bounds are a terrible idea.

PR #137, opened 2015-02-03
[Proposal] validCoupon method

lostincode:
Allows validation of coupon from Stripe.

### DIFF
--- a/src/Laravel/Cashier/StripeGateway.php
+++ b/src/Laravel/Cashier/StripeGateway.php
@@ -4,6 +4,7 @@ use Stripe;
 use Carbon\Carbon;
 use Stripe_Invoice;
 use Stripe_Customer;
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this is really a great solution because the coupon might have been used to many times, etc.

lostincode:
@taylorotwell from the docs:

https://stripe.com/docs/api?lang=php#coupon_object

````
valid (boolean)
Taking account of the above properties, whether this coupon can still be applied to a customer
```

tlaverdure:
@taylorotwell I think that this should be reviewed again. Not sure if this was always the case, but you can now supply a total number of times that a coupon can be used on Stripe. If the coupon is redeemed up to that amount, the `valid` property in the coupon object should return false.

PR #138, opened 2015-02-04
[Proposal] invoicePath method

DraperStudio:
I haven't found any way to get the path for an invoice so with the `invoicePath` method the local path for an invoice can be retrieved which then can be used to attach the invoice to an email or whatever else needs to be done with it.

```php
$invoice = $user->invoicePath($id, [
    'vendor'  => env('SERVICE_NAME'),
    'product' => $plan['name'],
]);

$billingMailer->send($user, $invoice);
```


taylorotwell:
Added `invoiceFile` method to 3.0 branch.

PR #140, opened 2015-02-05
Update CashierServiceProvider.php

toniperic:
Since Illuminate\Support\ServiceProvider method loadViewsFrom($path, $namespace) accepts first the path and then the namespace, shouldn't these arguments switch places?


toniperic:
Didn't see that the same PR was already submitted a few hours before me.

PR #141, opened 2015-02-10
Fix Docblock.

ChrisReid:
Incorrect parameter name.


taylorotwell:
Fixed on 3.0 branch which is the current release. Thanks.

PR #144, opened 2015-03-02
Creating new card API change

zslabs:
Stripe API change: https://stripe.com/docs/upgrades#2015-02-18

There is still an issue with setting (and getting) the last four digits, as the way you retrieve a card is much different than before https://stripe.com/docs/api#retrieve_card

Relevant Stack Overflow issue: http://stackoverflow.com/questions/28735182/laravel-cashier-trouble-updating-card-fatal-error?noredirect=1#comment45864842_28735182

### DIFF
--- a/src/Laravel/Cashier/StripeGateway.php
+++ b/src/Laravel/Cashier/StripeGateway.php
@@ -451,7 +451,7 @@ class StripeGateway
     {
         $customer = $this->getStripeCustomer();
 
-        $card = $customer->cards->create(['card' => $token]);
+        $card = $customer->sources->create(['card' => $token]);
 
         $customer->default_card = $card->id;
 
--- a/tests/StripeGatewayTest.php
+++ b/tests/StripeGatewayTest.php
@@ -177,8 +177,8 @@ class StripeGatewayTest extends PHPUnit_Framework_TestCase {
 		$gateway->shouldReceive('getStripeCustomer')->andReturn($customer = m::mock('StdClass'));
 		$gateway->shouldReceive('getLastFourCardDigits')->once()->andReturn('1111');
 		$customer->subscription = (object) ['plan' => (object) ['id' => 1]];
-		$customer->cards = m::mock('StdClass');
-		$customer->cards->shouldReceive('create')->once()->with(['card' => 'token'])->andReturn($card = m::mock('StdClass'));
+		$customer->sources = m::mock('StdClass');
+		$customer->sources->shouldReceive('create')->once()->with(['card' => 'token'])->andReturn($card = m::mock('StdClass'));
 		$card->id = 'card_id';
 		$customer->shouldReceive('save')->once();
 


taylorotwell:
Tests are failing.

taylorotwell:
Also you can set the API version your application is using.

zslabs:
Hey Taylor,
Tests updated and thanks for mentioning manually setting the API version my app will use. The tricky part for me becomes how do we manage the difference between what we set Stripe's API to use compared to when Cashier is updated? I'd assume we'd always want to keep Cashier using the latest and greatest from Stripe, but it seems like it'd become a bit messy with the breaking changes that occurred in that last change in the API. Thanks!

taylorotwell:
Yeah Stripe API isn't exactly the most stable thing I've seen. We probably need a good, easy to find version of the Stripe API that Cashier is currently supporting. For example, if we merge this PR now, do we break people when they upgrade Cashier if they are on an old Stripe API version?

zslabs:
Yep, while this PR fixes an issue with the new Stripe API removing the card attribute, additional work needs to be done for saving and retrieving the last four of the CC on file (access methods are different).

I'd love to hear other's thoughts or typical use-cases of staying on older versions of the library, but I'd also lean on of how you normally handle that in the L framework for all its dependencies. If Cashier is going to support a certain Stripe API version, then maybe Cashier should be setting this directly to avoid confusion.

taylorotwell:
I'm tempted to make this a major release of Cashier (4.0), I don't see any other really decent way to do this?

zslabs:
Sure, I think that'd make sense. So then are you in agreement about both updating to the latest API and Cashier manually setting the API version it is compatible with to rule out confusion? Thanks for the continued back 'n forths.

taylorotwell:
I think we should just document in Laravel docs which versions support which versions of the Stripe API.

zslabs:
Definitely. Since the Stripe API can be a bit flakey when it comes to consistency across updates, being as transparent as possible about version support is a great idea.

rossedman:
I am experiencing the same problems. They are upgrading their API faster than I can build my billing portal. Could this be something that is just set in `config/services`?

taylorotwell:
This will have to be done on a major branch.

taylorotwell:
I've created a 4.0 branch for handling apps on Stripe API 2-18 and later. I've made this change on that branch. What do we need to do for the last four digits changes?

zslabs:
Sounds good, thanks a lot @taylorotwell. So as far as the other changes, the way you get and set a card has completely changed, as it's not returned with the customer object anymore https://stripe.com/docs/api#retrieve_card I'm not sure if that means we'll not only need to store a customer_id locally, but possibly a card_id as well.

chillichicken:
Why can't you just drive this via composer ?
A certain cashier version works with a certain Stripe version, which will be updated accordingly.

PR #147, opened 2015-03-13
Update Invoice.php

loduis:
After finding a bill that has not been able to collect, it is appropriate to invoke the method **pay** in one step

https://github.com/laravel/cashier/blob/3.0/src/Laravel/Cashier/StripeGateway.php#L245


taylorotwell:
Just use the getStripeInvoice method.

PR #150, opened 2015-03-24
[Proposal] Change cancellation webhook to correct hook from Stripe API

TheShiftExchange:
So I'm not sure if this is because of the latest Stripe API changes - or if it has always been this way - but the webhook that Cashier is currently using the handle cancelled subscriptions is not ideal.

The problem is Cashier is currently looking for a `invoice.payment_failed` webhook, which occurs anytime a payment fails. Cashier then manually checks if the payment fails 3 times, then Cashier cancels the subscription *and assumes that Stripe has also cancelled the subscription*.

The problem is there are instances where Stripe could cancel a subscription at other times, or you configure Stripe to cancel after X attempts, not necessarily 3.

It makes more sense to default the webhook to `customer.subscription.deleted`. This way - for any reason - if Stripe cancels a subscription - then that is reflected in our application - regardless of the circumstances.

### DIFF
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -50,33 +50,20 @@ class WebhookController extends Controller
     }
 
     /**
-     * Handle a failed payment from a Stripe subscription.
+     * Handle a cancelled customer from a Stripe subscription.
      *
      * @param  array  $payload
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    protected function handleInvoicePaymentFailed(array $payload)
+    protected function handleCustomerSubscriptionDeleted(array $payload)
     {
-        if ($this->tooManyFailedPayments($payload)) {
-            $billable = $this->getBillable($payload['data']['object']['customer']);
+        $billable = $this->getBillable($payload['data']['object']['customer']);
 
-            if ($billable) {
-                $billable->subscription()->cancel();
-            }
+        if ($billable) {
+            $billable->subscription()->cancel();
         }
-
+        
         return new Response('Webhook Handled', 200);
-    }
-
-    /**
-     * Determine if the invoice has too many failed attempts.
-     *
-     * @param  array  $payload
-     * @return bool
-     */
-    protected function tooManyFailedPayments(array $payload)
-    {
-        return $payload['data']['object']['attempt_count'] > 3;
     }
 
     /**


TheShiftExchange:
Thinking outloud - this is a breaking change.

Maybe we should leave the function `tooManyFailedPayments(array $payload)` in, and that would mean this is a non-BC change?

The only thing is the webhook will be called twice for some people - for `invoice.payment_failed` and `customer.subscription.deleted` - but that shouldnt matter?

taylorotwell:
Hmm, yeah this is definitely better. How would it be a breaking change?

src/Laravel/Cashier/WebhookController.php
```diff
@@ -50,36 +50,23 @@ protected function eventExistsOnStripe($id)
...
+        $billable = $this->getBillable($payload['data']['object']['customer']);
 
-            if ($billable) {
-                $billable->subscription()->cancel();
-            }
+        if ($billable) {
+            $billable->subscription()->cancel();
         }
-
+
```
GrahamCampbell: cs

TheShiftExchange:
@taylorotwell - because the removal of the `tooManyFailedPayments()` function. If someone has overloaded the Controller and is calling that function manually themselves - if we remove it, it might break applications.

I think I should modify the PR and just leave the function there - although it wont be used - but is available for legacy application support?

Also - the other thing is this line of code:

```
if ($billable) {
       $billable->subscription()->cancel();
 }
```

I think we should change it to:

```
if (($billable) && ($billable->subscribed())) {
           $billable->subscription()->cancel();
}
```

What I discovered tonight is that if you try to run `$billable->subscription()->cancel()` on a user who does not have a subscription - then PHP throws an error that you are running `cancel()` on `null`.

Putting the extra check there ensure this never occurs. The reason it will start occuring after this change is legacy code might have them cancel the subscription in the original `handleInvoicePaymentFailed` function - and then it will be hit a second time in the new function when stripe calls the extra web hook.

So we are just putting contingencies in place to help prevent breaks in any current application as best as possible...

p.s. perhaps make this at least a 4.1 change - rather than a 4.0.x change?

taylorotwell:
I think your idea of just leaving the tooManyFailedPayments in there is fine. Just add a note to doc block that the method is deprecated and not being used anymore and will be removed in a future major version. It's fine to add the subscription() check too as well.

TheShiftExchange:
@GrahamCampbell - is there a convention to "deprecated" comments in doc block - or you happy if I just write something there?

GrahamCampbell:
@TheShiftExchange Could you use the `@deprecated` annotation please. Formatting it how symfony do should be good enough. Taylor will let you know if he doesn't like it. :)

TheShiftExchange:
Cool - thanks - I'll look that up now and do a new PR to conform to all the requests above

PR #151, opened 2015-03-25
[Proposal] Change cancellation webhook to correct hook from Stripe API

TheShiftExchange:
As discussed in closed PR https://github.com/laravel/cashier/pull/150

There is corresponding doc changes here as well: https://github.com/laravel/docs/pull/1277

### DIFF
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -50,19 +50,17 @@ class WebhookController extends Controller
     }
 
     /**
-     * Handle a failed payment from a Stripe subscription.
+     * Handle a cancelled customer from a Stripe subscription.
      *
      * @param  array  $payload
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    protected function handleInvoicePaymentFailed(array $payload)
+    protected function handleCustomerSubscriptionDeleted(array $payload)
     {
-        if ($this->tooManyFailedPayments($payload)) {
-            $billable = $this->getBillable($payload['data']['object']['customer']);
+        $billable = $this->getBillable($payload['data']['object']['customer']);
 
-            if ($billable) {
-                $billable->subscription()->cancel();
-            }
+        if (($billable) && ($billable->subscribed())) {
+            $billable->subscription()->cancel();
         }
 
         return new Response('Webhook Handled', 200);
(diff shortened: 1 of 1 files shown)


src/Laravel/Cashier/WebhookController.php
```diff
@@ -71,6 +69,7 @@ protected function handleInvoicePaymentFailed(array $payload)
     /**
      * Determine if the invoice has too many failed attempts.
      *
+     * @deprecated Use Stripe webhook 'customer.subscription.deleted' instead
```
GrahamCampbell: This annotation should be separated with an extra line after it, and the description should end in a fullstop.

taylorotwell:
So just to confirm you have tested this with a real cancellation? :)

src/Laravel/Cashier/WebhookController.php
```diff
@@ -50,19 +50,17 @@ protected function eventExistsOnStripe($id)
...
+    protected function handleCustomerSubscriptionDeleted(array $payload)
     {
-        if ($this->tooManyFailedPayments($payload)) {
-            $billable = $this->getBillable($payload['data']['object']['customer']);
+        $billable = $this->getBillable($payload['data']['object']['customer']);
 
-            if ($billable) {
-                $billable->subscription()->cancel();
-            }
+        if (($billable) && ($billable->subscribed())) {
```
taylorotwell: You don't need parens around either of these variables / expressions.

TheShiftExchange:
I've tested it with the Stripe Test API.

If you want to leave this PR for a few days - I'm happy to try this fork on my app and confirm.

taylorotwell:
I'll go ahead and merge it and I can test it too because I actually have a little Cashier test application (should probably open source that - heh). Once we've tested it I'll tag it.

taylorotwell:
Yeah this is at least partially broken. Firstly because since the subscription has been deleted from Stripe. Trying to delete it again will ALWAYS break.

TheShiftExchange:
Because `cancel()` is calling back to Stripe and re-telling it to cancel the subscription?

Hmmm... so is there a command in Cashier that only cancels a subscription on the Cashier side - without a call back to Stripe?

i.e.

```
if (($billable) && ($billable->subscribed())) {
         $billable->subscription()->cancelOnlyForCashier();
}
```

taylorotwell:
I fixed it in a recent commit.

On Wed, Mar 25, 2015 at 10:49 AM, Laurence Ioannou <notifications@github.com

TheShiftExchange:
Cool - your worth more money... :)

PR #152, opened 2015-04-06
updated to stripe/stripe-php ~2.0

ellisthedev:
Hello,

I'm the one who reached out to you on Twitter asking about this. Figured I'd lend a hand :) It appears this project does not have an Issues tab or I would have created the `[Proposal]` issue first. I have upgraded the project to `stripe/strip-php=~2.0` and made the necessary class name changes.

Running `phpunit` locally with this branch does not return any errors.

There is also one additional change that was made, which can be viewed [here](https://github.com/laravel/cashier/compare/4.0...ellisthedev:stripe-2.0#diff-b0e465721b7c0c823d5e9a9164e51294L174). This change was to fix an error that was being returned during the php unit testing. The `money_format` method was returning `USD100.00` and then that being passed into `round()` was causing the function to return `0`; thus failing [this condition](https://github.com/laravel/cashier/blob/4.0/tests/InvoiceTest.php#L72).

If you need me to do anything else please let me know.

Cheers,
Andrew

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "illuminate/filesystem": "~5.0",
         "illuminate/support": "~5.0",
         "nesbot/carbon": "~1.0",
-        "stripe/stripe-php": "~1.9",
+        "stripe/stripe-php": "~2.0",
         "symfony/http-kernel": "~2.3"
     },
     "require-dev": {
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
This is breaking, so absolutely can't go to the 2.0 branch. Please close this and send to the master (5.0) branch.

ellisthedev:
Hey @GrahamCampbell sorry, the title is confusing. The update is updating the dependency `stripe/stripe-php=~1.9` to `stripe/stripe-php=~2.0`. I will reopen this on the `master` branch though. Uno momento.

PR #153, opened 2015-04-07
Updating to stripe/stripe-php=~2.0

ellisthedev:
Reopening on the `master` branch. Refer to the prior pull request for details: https://github.com/laravel/cashier/pull/152

### DIFF
--- a/src/Laravel/Cashier/Invoice.php
+++ b/src/Laravel/Cashier/Invoice.php
@@ -171,7 +171,7 @@ class Invoice
     {
         setlocale(LC_MONETARY, $this->billable->getCurrencyLocale());
 
-        return round(money_format('%i', ($this->subtotal / 100) - ($this->total / 100)), 2);
+        return round(money_format('%!i', ($this->subtotal / 100) - ($this->total / 100)), 2);
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "illuminate/filesystem": "~5.0",
         "illuminate/support": "~5.0",
         "nesbot/carbon": "~1.0",
-        "stripe/stripe-php": "~1.9",
+        "stripe/stripe-php": "~2.0",
         "symfony/http-kernel": "~2.3"
     },
     "require-dev": {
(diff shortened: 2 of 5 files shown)


ellisthedev:
This version may be breaking now that I look at the fact you allow passing `$apiKey` to methods when the parent class expects an array of options. However, this part is untested. I'm doing a lot of work with subscriptions and one off charges this week so I will test this code in depth and report back if this needs more work.

@taylorotwell also notified me on Twitter that he has a test Stripe application that he's going to test this with as well. So if he finds anything himself I'm sure we'll hear about it.

src/Laravel/Cashier/Invoice.php
```diff
@@ -171,7 +171,7 @@ public function discount()
     {
         setlocale(LC_MONETARY, $this->billable->getCurrencyLocale());
 
-        return round(money_format('%i', ($this->subtotal / 100) - ($this->total / 100)), 2);
+        return round(money_format('%!i', ($this->subtotal / 100) - ($this->total / 100)), 2);
```
taylorotwell: What does this do?

ellisthedev: So this change was to fix an error that was occurring on the unit tests. The original method the `money_format` call was returning `USD100.00`. Then when passed into round, round was returning `0`; thus breaking the unit test as `0 !== 100`.

Adding the bang keeps money_format from returning `USD` in the string.

SylarRuby:
@taylorotwell after I ran composer update Im getting this for new users to my website: Could not determine which URL to request: Laravel\Cashier\Customer instance has invalid ID: Was there an upgrade?

taylorotwell:
Which version of cashier are you on. This hasn't been merged yet.

SylarRuby:
Hey @taylorotwell It a pleasure lol. The when I place this in my controller: $invoices = $user->invoices(); I get the error when a new user registers.

ellisthedev:
This sounds like a bug in your application. That error is you trying to call Stripe with a cus_ ID that does not exist in Stripe.

Have you verified that ID exists?

SylarRuby:
Ahhhh but how to call that method? It works ok but if a new user registers in my website it throws an error. I cant even wrap it around an if statement.

PR #154, opened 2015-04-14
Update Invoice.php

sorora:
Checking whether an invoice has a discount is inaccurate when an invoice has tax applied as the sub total and total do not match. This will resolve the issue, or a check for if the subtotal + tax equals the total would work too.

### DIFF
--- a/src/Laravel/Cashier/Invoice.php
+++ b/src/Laravel/Cashier/Invoice.php
@@ -139,7 +139,7 @@ class Invoice
      */
     public function hasDiscount()
     {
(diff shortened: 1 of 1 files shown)


sorora:
Has this failed just because of missing env files?

taylorotwell:
How about leaving the other check in there as well as adding the null discount check? Also you would probably want to send this to the default branch instead of master.

sorora:
Done at #155 - thanks for all the great work on this, I've learnt quite a bit just from the source!

PR #159, opened 2015-04-28
Added tax_percent to subscription

matism:
A user can set the tax_percent parameter directly while creating a customer:

```
$user->subscription('plan',20.0)->create($token);
```


taylorotwell:
This was recently done by another PR.

PR #161, opened 2015-05-06
Allow testing of Webhooks

T3chn0crat:
The issue is that you cannot test Stripe Webhooks because of the brute force protection.  The test event from Stripe.com has an id of evt_00000000000000 which obviously cannot be validated.  This makes it impossible to test.  

So if in debug mode allow this protection to be bypassed.  Or maybe add in a services.stripe.testing or something like that.

### DIFF
--- a/feajvkhr.ict.txt
+++ b/feajvkhr.ict.txt
@@ -1,0 +1,11 @@
+﻿Squash
+
+# Please enter the commit message for your changes. Lines starting
+# with '#' will be ignored, and an empty message aborts the commit.
+# On branch patch-1
+# Your branch is up-to-date with 'origin/patch-1'.
+#
+# Changes to be committed:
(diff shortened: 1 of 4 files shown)


T3chn0crat:
I think that would be a fine way to implement it as well.  Bottom line is the way it is right now it will not allow testing.

Why is my change breaking?

TheShiftExchange:
Well - maybe "breaking" is an overstatement - but it would change the behavior of the function in a debug environment. So having its own config means you keep the original functionality, and provide a choice.

T3chn0crat:
Sorry about all the mess @taylorotwell it was my first time working with Travis and all the automated tests.  Hopefully this works for you.

GrahamCampbell:
Please squash to one commit.

T3chn0crat:
Well I have made this even worse.  I will redo it and resubmit the pull request

PR #162, opened 2015-05-08
Fix the ability to test webhooks

T3chn0crat:


taylorotwell:
We can't use a facade from a package like this.

T3chn0crat:
Ok.  I just tried to follow how you did it else where in Cashier.  Then what is the proper course of action in your opinion?  Injecting it?

The test event from Stripe.com has an id of evt_00000000000000 which obviously cannot be validated. As it sits right now the brute force protection makes it impossible to test with out deleting the protection, which should not be the way this works.  Also since there is no response from the handler when it kicks it out, anyone else trying it will be in the same boat I was where you don't know why the tests are not working.

PR #164, opened 2015-05-12
Fixing tax_percent issue

nikospkrk:
Stripe throwns an error when tax_percent is empty ( Invalid decimal: ; must contain at maximum two decimal places ), let's default it to 0.


taylorotwell:
I tweaked this to just conditoinally add the tax_percent if it is actually set.

taylorotwell:
Tagged 5.0.3.

PR #165, opened 2015-05-14
Fixing an issue with not being able to test stripe webhooks

christopherrolfe198:
I thought this was closer to being a bug that being a feature so I haven't created an issue for this.

When trying to test webhooks through the stripe interface my requests were being blanked because the events don't exist on stripe's end.

I've added a minor change that checks if we are in live mode and continues if we aren't.

### DIFF
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -21,7 +21,7 @@ class WebhookController extends Controller
     {
         $payload = $this->getJsonPayload();
 
-        if (! $this->eventExistsOnStripe($payload['id'])) {
+        if (! $this->eventExistsOnStripe($payload['id']) and $payload['livemode']) {
             return;
         }
 


taylorotwell:
But wouldn't you sometimes actually want to test events in test mode?

christopherrolfe198:
Interesting point, I don't think this would stop that working. I'll need to look into it to be certain.
The check I've added simply says if the event does exist on stripe and we are in livemode return nothing.
Whereas if the event does exist and we are in test mode it continues.

christopherrolfe198:
So I've just done a test with creating a plan and the request reached my handler so it works as mentioned before.

If the event doesn't exist on stripe and you are in livemode it returns early. Otherwise it will continue on to the appropriate handle function.

T3chn0crat:
@taylorotwell - Sorry but I do not think this is a proper fix for this problem.  Though it helps by now allowing testing, you are invalidating the brute force protection that was added now by simply adding a livemode to the payload.  I don't think that's what you want to do. 

My pull #162 at least made it so you had to actually enable it to bypass the protection.

PR #166, opened 2015-05-19
Change onPlan method

jpavon:
Is there any reason why this method is not like this? Maybe create other method named subscribedOnPlan()?


taylorotwell:
It seems to work as is?

jpavon:
I was trying to check when a user is on a plan, but when they are on grace period and on trial this method returns false. 

Also I'm not sure how $this->subscription()->planId() works, but it seems is checking to stripe directly, when a  check to the database is enough?

PR #178, opened 2015-07-02
add support to get the upcoming invoice

kyranb:
branches 3.0 to 5.0 have it so why not 2.0 as well :+1:


taylorotwell:
2.0 is an old version. I would upgrade to 3.0.

kyranb:
Unfortunately it's for an app that's still running Laravel 4.2 which Cashier 3.0 does not support.

PR #184, opened 2015-07-09
newQuantity

TheShiftExchange:
There is no easy way to set a new Quantity on a subscription. Currently the only choices are `increment` and `decrement` - but that assumes you want to only increase/decrease from a previous known amount. 

There are useful times where you just want to set a specific quantity - regardless of any previous number.

Before the pull you could do this:

```
$user->subscription()->updateQuntity($user->subscription()->getStripeCustomer(), 10);
```

Now you can just do

```
$user->subscription()->newQuantity(10);
```


taylorotwell:
I tweaked this so you can just use `updateQuantity`.

PR #188, opened 2015-07-28
Return 200 response on failed eventExistsOnStripe

T3chn0crat:
This will fix a bit of an edge case.

It appears that Stripe sends webhooks to one site at a time that are listed in your webhooks list. If any of the sites fail it will not move on to the next. Any response other than a HTTP 2xx message is considered a fail. Thus this will stop that from happening.

### DIFF
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -21,7 +21,7 @@ class WebhookController extends Controller
         $payload = $this->getJsonPayload();
 
         if (! $this->eventExistsOnStripe($payload['id'])) {
-            return new Response('Webhook Handled', 200);
+            return;
         }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
An empty return already will return a 200 empty response.

T3chn0crat:
@taylorotwell Weird your right.  Then I am stumped by why Stripe stopped processing the webhooks for until I passed it back.  I guess more testing.

Can we at least include the pull so we know that its hitting the handler?  That would be helpful in tracking down issues rather than a blank message.

T3chn0crat:
@taylorotwell Sorry, I reverted back the commit when you closed it.  So you merged in the revert vs the pull.  At least I am assuming you meant to pull in my commit.  Sorry about that.

PR #190, opened 2015-08-08
Allow testing of Webhooks

T3chn0crat:
This will allow the ability to bypass the eventExistsOnStripe protection if services.stripe.test exists and is set to true.  Without either it will fall back to the indented protection.  So if someone doesn't want to test, they do not have to worry about this.

With out this it is difficult to test webhooks because it obviously doesn't have a valid event id.  I have to constantly go in and comment out the return to test.

Also added Webhook Handled to the return so it's easier to see if the hook is at least getting that far, which makes it easier to debug.

### DIFF
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -21,8 +21,10 @@ class WebhookController extends Controller
     {
         $payload = $this->getJsonPayload();
 
-        if (! $this->eventExistsOnStripe($payload['id'])) {
-            return;
+        if (! class_exists('Config') || ! Config::has('services.stripe.test') || ! Config::get('services.stripe.test')) {
+            if (! $this->eventExistsOnStripe($payload['id'])) {
+                return new Response('Webhook Handled', 200);
+            }
         }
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This was already proposed and rejected, sorry.

T3chn0crat:
Last time it was rejected #162 because of the way I tried to pull in Config, I believe.  

I forgot that test needs Config to pass.  If it would be better I can push this to the eventExistsOnStripe method.

This annoys me very time there is an update.  I have to go in a manually comment this out to pass my tests.

T3chn0crat:
@taylorotwell I got this to pass using the following:

Move the Config checks to the eventExistsOnStripe method however that did not seem like the correct implementation to me.

Explicitly using Illuminate\Support\Facades\Config; however that's what bounced #165  

Finally by adding in class_exists for Config.  Which is what I settled on, but it seems like not the cleanest way.

taylorotwell:
I agree this needs to be fixed but still not sure this is correct way. What about just using a simple $_ENV variable? That would work well with `.env` files and doesn't require any kind of Facade usage.

PR #191, opened 2015-08-10
Load StripeGateway from the container

phleobon:
Loading the Laravel\Cashier\StripeGateway from the containers allows to mock it in functional test.

While trying to test the user registration + subscription process, you might not want to hit the stripe API.

### DIFF
--- a/src/Laravel/Cashier/Billable.php
+++ b/src/Laravel/Cashier/Billable.php
@@ -55,7 +55,7 @@ trait Billable
      */
     public function subscription($plan = null)
     {
-        return new StripeGateway($this, $plan);
+        return app('Laravel\Cashier\StripeGateway', [$this, $plan]);
     }
 
     /**


src/Laravel/Cashier/Billable.php
```diff
@@ -55,7 +55,7 @@ public function charge($amount, array $options = [])
      */
     public function subscription($plan = null)
     {
-        return new StripeGateway($this, $plan);
+        return app('Laravel\Cashier\StripeGateway', [$this, $plan]);
```
GrahamCampbell: You can't use the app helper because it's not available in non-laravel app.s

GrahamCampbell: Cashier does not depend on laravel.

phleobon:
Oh, I wasn't aware about this.

phleobon:
By the way, any advice on how I could mock it then ?

GrahamCampbell:
Instance mocking perhapse?

phleobon:
I'll try that. Thanks

PR #193, opened 2015-08-11
Allow testing of Webhooks using ENV

phleobon:
After seeing the pull request #190 I wanted to contribute with what looks to me a better solution.

As @taylorotwell said using environment variables looked like a better choice.

To keep it simple I added a isInTestingMode method that return a boolean according to the CASHIER_TESTING env variable.

Then in the handleWebhook method we can verify that we are not in testing mode.

### DIFF
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -50,7 +50,7 @@ class WebhookController extends Controller
     }
 
     /**
-     * Check if testing mode is on
+     * Verify if cashier is in testing mode.
      * 
      * @return bool
      */
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -21,7 +21,7 @@ class WebhookController extends Controller
     {
         $payload = $this->getJsonPayload();
 
-        if (! $this->eventExistsOnStripe($payload['id'])) {
+        if (! $this->eventExistsOnStripe($payload['id']) && ! $this->isInTestingMode()) {
             return;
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Build still failing.

src/Laravel/Cashier/WebhookController.php
```diff
@@ -50,6 +50,16 @@ protected function eventExistsOnStripe($id)
     }
 
     /**
+     * Check if testing mode is on
+     * 
+     * @return bool
+     */
+    protected function isInTestingMode()
+    {
+        return (bool) getenv('CASHIER_TESTING');
```
GrahamCampbell: Maybe we should check if APP_ENV is equal to testing instead of adding yet another variable.

src/Laravel/Cashier/WebhookController.php
```diff
@@ -50,6 +50,16 @@ protected function eventExistsOnStripe($id)
     }
 
     /**
+     * Check if testing mode is on
```
GrahamCampbell: This doesn't read well

src/Laravel/Cashier/WebhookController.php
phleobon: I have thought about this, but when testing you might not always put the APP_ENV to testing.

But in case you have multiple testing frameworks with different APP_ENV then you won't be in test mode.

Adding a CASHIER_TESTING variable allows to put in in testing without having to care about the environment you are using.

taylorotwell:
Changed it to `cashier_env` which should be set to `testing`.

PR #194, opened 2015-08-14
Add livemode parameter for testing stripe webook in test mode

Steingate:
Hello,

Actually is not possible to test webhook in test mode (stripe), so i've had an another check of livemode to the handleWebhook method:
```php
if (!$this->eventExistsOnStripe($payload['id']) && $payload['livemode']) {
    return;
}
```

Because in test mode, stripe always send fake event id, so eventExistsOnStripe always return false.
With this, we can test webhook in testmode (because we dont care if the event exist in test mode).

PS: Test mode: livemode = false.
       Livemode: livemode = true

what do you think ?

Alexandre

### DIFF
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -21,7 +21,7 @@ class WebhookController extends Controller
     {
         $payload = $this->getJsonPayload();
 
-        if (!$this->eventExistsOnStripe($payload['id']) && $payload['livemode']) {
+        if (! $this->eventExistsOnStripe($payload['id']) && $payload['livemode']) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/Laravel/Cashier/WebhookController.php
```diff
@@ -21,7 +21,7 @@ public function handleWebhook()
     {
         $payload = $this->getJsonPayload();
 
-        if (! $this->eventExistsOnStripe($payload['id'])) {
+        if (!$this->eventExistsOnStripe($payload['id']) && $payload['livemode']) {
```
GrahamCampbell: don't remove that space

Steingate: Done, thx

taylorotwell:
Couldn't someone just craft a false payload to not include the `livemode` variable?

Steingate:
I'm not sur a complety understand what you want to said, Stripe Test webhook always send a payload with a livemode key. This key is the only method to know if the webhook was send in a test or a live mode.

How to test my webhook without a fix like that?

Thx

PR #198, opened 2015-08-23
Added deleted check for getStripeCustomer

JTallis:
Added a check to make sure customers are not deleted before proceeding to find an existing subscription. `Customer::usingMultipleSubscriptionApi()` will cause an error as Customer::subscriptions does not exist for deleted models. Deleted customers do not have existing subscriptions so proceeding to find a subscription is not necessary.

The error is Stripe Notice: Undefined property of Laravel\Cashier\Customer instance: subscriptions

### DIFF
--- a/src/Laravel/Cashier/StripeGateway.php
+++ b/src/Laravel/Cashier/StripeGateway.php
@@ -551,7 +551,7 @@ class StripeGateway
     {
         $customer = Customer::retrieve($id ?: $this->billable->getStripeId(), $this->getStripeKey());
 
-        if ($customer->deleted === false && $this->usingMultipleSubscriptionApi($customer)) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Checking customer->deleted will throw an exception if the csutomer is not deleted and you try to access that property. You need to use `isset`

Source: tested it.

JTallis:
Oh, shoot. You're right. The exception thrown is from Object.php:93 as a result of returning null instead of a variable reference. I changed this method myself to match that of Stripe PHP 3.2 so it returns null without an error and I didn't realise customer->deleted did not exist.

taylorotwell:
As you can see in StyleCI build, it wants you to add a space after the `!`.

PR #206, opened 2015-09-29
fixes two bugs

beeseenmedia:
please see edits.

fixes cancellation problem when you cancel at end of billing period and then cancel it immediately.

fixes trial period bug

### DIFF
--- a/src/Laravel/Cashier/WebhookController.php
+++ b/src/Laravel/Cashier/WebhookController.php
@@ -66,6 +66,12 @@ class WebhookController extends Controller
         return new Response('Webhook Handled', 200);
     }
     
+    /**
+     * Handle a failed payment.
+     *
+     * @param  array  $payload
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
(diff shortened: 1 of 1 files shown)


taylorotwell:
Needs much more description. How to recreate, explanation of fixes made, etc.

beeseenmedia:
@taylorotwell 

1.  I added a Invoice Payment Failed webhook.  There was an issue with the trial.  The trial would be for x amount of days After, stripe would attempt to charge the credit card (even if there was none), it sends the webhook,  then two days later it would try again.  It attempts 3 times to charge the credit card before sending the cancelation webhook, which meant the trial was extended to x + days attempting to charge.  I added the Payment Failed webook which basically says, if payment fails and there is no credit card on file then cancel immediately.

2.  cancel subscription has a bug in which if you cancel(true) it sets stripe active = 0 and subscription ends = future, if you then cancel immediately it fails to set subscription end = now because the user is no longer subscribed so it bypasses the if ($customer->subscription).  by moving the line down it makes sure that it sets subscription ends = now

PR #213, opened 2015-10-08
Simple refactory

edueo:
Remove unnecessary else.


taylorotwell:
We don't remove these.

PR #228, opened 2015-11-26
Use more consistent lowerbound on php

GrahamCampbell:


taylorotwell:
Heh saw this after I did it

PR #249, opened 2016-01-14
Set foreign key for subscriptions() relation

dsshafer:
Set the foreign key for the subscriptions() relation to "user_id". The user() relation in Subscription.php already sets the local key to "user_id", so this is just the other side of the relationship. This is necessary to handle cases where the Stripe model isn't "User".

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -102,7 +102,7 @@ trait Billable
      */
     public function subscriptions()
     {
-        return $this->hasMany(Subscription::class)->orderBy('created_at', 'desc');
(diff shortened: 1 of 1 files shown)


taylorotwell:
I would just override that method in your own model if you need to.

dsshafer:
I guess the question is: Do you want Cashier to guess the foreign key for the Billable->Subscription relationship, or do you want it always to assume 'user_id'? Because right now it's inconsistent. On the Subscription->user() side, it always assumes 'user_id'. But on the Billable->subscriptions() side, it guesses based on the model name. I just think it should be consistent.

PR #252, opened 2016-01-14
[6.0] Fix docblock

lucasmichot:
Fix docblock


lucasmichot:
@taylorotwell seems the Stripe sandbox API key has not been set in Travis ENV variables
Resulting the tests to fail

PR #259, opened 2016-01-25
Update package name to match packagist existing name

phazei:
Currently the package.json name was changed to "new-cashier".  This doesn't match the packagist name.  From what I could find out, there's not a simple way to update the packagist name.
Because the names don't match up, it makes it not possible to fork the repo and override it in the "repositories" root object.


GrahamCampbell:
lol

PR #260, opened 2016-01-26
StripeCharge::create() allows for both params and options arguments, …

dmcfaul:
…see: https://github.com/stripe/stripe-php/blob/master/lib/Charge.php#L29-L38


GrahamCampbell:
This is a breaking changes.

dmcfaul:
Could you elaborate on why, I'm not sure I see the breakage? I did swap the params and options arg names to avoid confusion with the stripe libraries' implementation, however I updated the functions to reflect that change.

Issue #261, opened 2016-01-28
[Proposal] Don't check event existence with Stripe servers if the result is neglected


jeteon:
The line:
https://github.com/laravel/cashier/blob/6.0/src/Http/Controllers/WebhookController.php#L23
checks that events exist on Stripe. However, the check on the servers is ignored if the environment is currently `testing`.

This means that all tests pay the price of the check even though tests will disregard the result, which makes test unnecessarily slow. I propose simply reversing the operands in the `if` statement. This won't change the effective logic as far as I can see but will make tests faster.

PR #262, opened 2016-01-28
Change order of boolean test for faster tests

jeteon:
The logic of the Boolean expression in the previous order makes it so that the event existence is always checked against the Stripe servers first before checking whether this is a test. This results in slower tests as each test will make this round trip even though the condition will make it so the result is irrelevant.

The proposed ordering omits the check against the Stripe servers entirely during tests. This change has no real consequence in terms of effective logic since even if one explicitly wanted it to check that the event exists in Stripe, the `&& !` was making the result irrelevant when testing.

This is the propose PR for #261.

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -20,7 +20,7 @@ class WebhookController extends Controller
     {
         $payload = json_decode($request->getContent(), true);
 
-        if ( ! $this->isInTestingEnvironment() && ! $this->eventExistsOnStripe($payload['id'])) {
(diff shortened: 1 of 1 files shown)


jeteon:
The build errors seem to be coming from elsewhere. I don't understand how the changes I made did that. I'll work on it though.

jeteon:
As far as I can tell `getenv('STRIPE_SECRET')` returns nothing in each case. It seems other PRs you've merged before also fail in the same way. I'm assuming this is intentionally done so you don't have Stripe credentials out in the open.

PR #264, opened 2016-01-31
Add relation on a given model, rather than hard-coded relation

nathanjdunn:
At the moment when you try to cancel a subscription on Cashier and are not using the standard User model, it fails since it is trying to look up the `user_id` relation in the DB when that may not exist. My fix allows Cashier to look up the relation from a config setting or guess based from the model.


taylorotwell:
Am I missing something or do you never use this variable?

nathanjdunn:
Nope, it would help if I did though! Will close PR and make a new one...

PR #267, opened 2016-02-02
Fixed typo's in Subscription.php

mitchjam:
Fixed minor typo's in _src/Subscription.php_.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -70,7 +70,7 @@ class Subscription extends Model
     }
 
     /**
-     * Determine if the subscription is within it's grace period after cancellation.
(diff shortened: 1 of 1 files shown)


src/Subscription.php
```diff
@@ -70,7 +70,7 @@ public function onTrial()
     }
 
     /**
-     * Determine if the subscription is within its grace period after cancellation.
+     * Determine if the subscription is within it's grace period after cancellation.
```
taylorotwell: This actually isn't correct :smile: ... it's is a contraction for "it is".

mitchjam: My apologies. Muscle memory :smile:.

Issue #268, opened 2016-02-07
Issue with using swap() when stripe model is something other than User


jakedewitt87:
Hi Guys,
   I just ran into an issue on one of my projects where I am using Cashier on Laravel 5.2, where I have users under an account model, and it is the account model that the subscriptions should connect to, so I updated my Stripe model in my config > services.php file to:

```
'stripe' => [
    'model'  => App\Account::class,
    'key'    => env('STRIPE_KEY'),
    'secret' => env('STRIPE_SECRET'),
],
```

 The issue is I am building out a change subscription plan section in my project, and I am using the subscription swap() method in cashier, but when I try to call this I am getting the following error:

FatalThrowableError in Subscription.php line 285: Fatal error: Call to a member function asStripeCustomer() on null

I think I tracked down the issue to this Subscription.php file, line 35:

 return $this->belongsTo($model, 'user_id'); 

In my subscriptions table I have named my foreign key 'account_id' instead of 'user_id'. Is this an issue with something I am doing or should this be updated to not hard code the user_id as the foreign key in the code? Thanks guys.

kyranb:
@taylorotwell Any chance of having this merged. I've been holding off a L4.2 upgrade due to this.

PR #269, opened 2016-02-09
[Proposal] Match subscription column name to class name

invento:
#268


taylorotwell:
This still doesn't solve that the method is named "users".

PR #271, opened 2016-02-13
Fix newSubscription with null tokens

abaxo:
Fix for https://github.com/laravel/cashier/issues/270 - attempting to update card with a null token

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -362,7 +362,9 @@ trait Billable
         // Next we will add the credit card to the user's account on Stripe using this
         // token that was provided to this method. This will allow us to bill users
         // when they subscribe to plans or we need to do one-off charges on them.
-        $this->updateCard($token);
+        if ($token !== null) {
+            $this->updateCard($token);
+        }
 
         return $customer;
     }


taylorotwell:
What is your situation of wanting to create a Stripe customer without a token?

abaxo:
The scenario is that I have a multi-tenant app where a customer is a company rather than an individual user. As part of our pre-sales activities, the sales team creates a tenant (company) in our software which in turn creates users and other items which are 'owned' by the tenant. The tenant's users then have access to the software on the plan which the pre-sales activities have identified as being most suitable for the tenant for a fixed period of time (the trial period). It is important that we don't force collection of billing details at this point, because at the time the customer is created we are still moving through the sales journey, but flexible enough that we can add the card details later to continue the subscription. 

It is also important that an individual user doesn't own the billing relationship because in an enterprise the billing will either be manual via a PO or via a company card, where multiple users will need to be able to own the billing relationship.

I had considered managing trials through the Models directly but we want/need Stripe to be the source of truth for subscription management.

PR #210, opened 2015-10-01
billing_cycle_anchor Support (on 5.0 branch)

ryross:
Here's the commit on 5.0.

See https://github.com/laravel/cashier/pull/209

Thanks!

PR #285, opened 2016-03-07
Added no prorate and billing cycle anchor setters for subscription swapping.

RyanHavoc:
**Moved this onto branch 6.0 as requested**

This is a little update that allows you to set prorate and billing_cycle_anchor when swapping a subscription plan.

Usage is:

$user->subscription()->noProrate()->billingCycleAnchor("now")->swap($stripe_plan);

I proposed this need way back in 2014 on this pull request: #88

It looks like it was implemented elsewhere: #210

But subsequently dropped in Cashier 6.0 (along with noProrate()) unless I'm massively overlooking something.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -188,7 +188,7 @@ class Subscription extends Model
      * @param string $date
      * @return $this
      */
-    public function billingCycleAnchor($date = "now")
+    public function billingCycleAnchor($date = 'now')
     {
         $this->billing_cycle_anchor = $date;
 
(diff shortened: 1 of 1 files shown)


matt-west:
:+1:

taylorotwell:
What is supposed to be passed into this billing cycle anchor method? A string? A UNIX timestamp? A DateTime object?

RyanHavoc:
@taylorotwell It's a Unix timestamp or special string value "now". Works the same way as trial_end. I've updated the docblock to clarify.

PR #293, opened 2016-04-20
Refunds invoices

mlantz:
Now you can refund a user

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -64,6 +64,7 @@ trait Billable
     public function refund($charge, array $options = [])
     {
         $options['charge'] = $charge;
+
         return StripeRefund::create($options, ['api_key' => $this->getStripeKey()]);
     }
 
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -53,7 +53,7 @@ trait Billable
     }
 
     /**
-     * Refund a customer for a charge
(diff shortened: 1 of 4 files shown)


src/Billable.php
```diff
@@ -184,13 +201,11 @@ public function invoice()
...
         if ($this->stripe_id) {
             try {
-                StripeInvoice::create(['customer' => $this->stripe_id], $this->getStripeKey())->pay();
+                return StripeInvoice::create(['customer' => $this->stripe_id], $this->getStripeKey())->pay();
             } catch (StripeErrorInvalidRequest $e) {
                 return false;
             }
         }
-
-        return true;
```
GrahamCampbell: this refactor is incorrect

mlantz: Suggestions? I'd like to be able to get back the invoice object from Stripe, and returning the object from Stripe would not be considered false.

mlantz:
Didn't realize what I had done. Better?

src/Billable.php
```diff
@@ -52,6 +53,22 @@ public function charge($amount, array $options = [])
     }
 
     /**
+     * Refund a customer for a charge.
+     *
+     * @param  str  $charge
```
GrahamCampbell: string, not str

PR #294, opened 2016-05-11
Add tax to single charge with invoice

michellaurent:
Add the tax percentage when create a new Stripe Invoice like for subscription

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -201,7 +201,10 @@ trait Billable
     {
         if ($this->stripe_id) {
             try {
-                return StripeInvoice::create(['customer' => $this->stripe_id], $this->getStripeKey())->pay();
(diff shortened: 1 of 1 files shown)


taylorotwell:
Right now I'm going to hold off on this because it makes the system less flexible if we are forcing the tax percentage on every invoice charge. There may be situations where you want to manually specify that percentage or not charge it at all.

michellaurent:
@taylorotwell I don't understand why it shouldn't work the same way for one time invoice and subscription... Couldn't we pass an optional tax rate in ``invoiceFor()`` or a ``$taxable`` boolean ?
Without this I feature I can't use Cashier and Spark for one time invoice in EU.

ultraN:
Still no taxable option?

PR #297, opened 2016-05-23
Add hasCardOnFile method to Billable

Francismori7:
Allows you to check if the customer currently has a card on file.

I couldn't get tests to run on my system.

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -69,6 +69,45 @@ trait Billable
     }
 
     /**
+     * Determines if the customer currently has a card on file.
+     *
+     * @return bool
+     */
+    public function hasCardOnFile()
+    {
+        // If we know already that there's a card on file, don't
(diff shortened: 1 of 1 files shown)


Francismori7:
Allows you to check if the customer currently has a card on file. I couldn't get tests to run on my system.

Do not merge yet. Some issues happened on my end.

taylorotwell:
What situations have you had where the local database doesn't have the card information but Stripe does have card information?

Francismori7:
While testing I would add the card in stripe and it wouldn't follow too good and caused issues. 

I was just checking card brand in the user data and figured a fallback option would be nice.

Again do not merge there's an issue with default source

taylorotwell:
OK make another PR when your'e ready.

Francismori7:
Will do tonight or something like that, got it fixed in my codebase alreary

PR #298, opened 2016-05-26
Add hasCardOnFile method to Billable

Francismori7:
Allows you to check if the customer currently has a card on file.

This PR adds three methods:
* `public function hasCardOnFile() : bool` **Determines if the user, according to the local DB, has a card linked.**
* `public function syncCustomerCardDetails : $this` **Forces the refresh of the customer's card details from Stripe's API.**
* `protected function fillCardDetails : $this` **Replaces the data about the card in the database for the Billable model.**

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -83,13 +83,20 @@ trait Billable
 
         $customer = $this->asStripeCustomer();
 
-        // If we determines that the customer has a card (desync?),
-        // we'll set it back onto the user model and save it.
-        if ($source = $customer->default_source) {
-            $this->fillCardDetails($source)->save();
+        $source = null;
+
+        foreach ($customer->sources->data as $card) {
+            if ($card->id === $customer->default_source) {
+                $source = $card;
+                continue;
+            }
         }
 
-        return (bool) $source;
+        // If we determines that the customer has a card (desync?),
+        // we'll set it back onto the user model and save it.
+        $this->fillCardDetails($source)->save();
+
+        return (bool)$source;
     }
 
     /**
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -69,6 +69,45 @@ trait Billable
     }
 
     /**
+     * Determines if the customer currently has a card on file.
+     *
+     * @return bool
+     */
+    public function hasCardOnFile()
+    {
+        // If we know already that there's a card on file, don't
+        // unnecessarily hit the Stripe API.
+        if ($this->card_brand !== null) {
+            return true;
+        }
+
+        $customer = $this->asStripeCustomer();
+
+        // If we determines that the customer has a card (desync?),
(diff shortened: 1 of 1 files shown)


taylorotwell:
So one thing that makes me a little hesitant here is the method name is "hasCardOnFile", which implies it's just going to return true / false. However, it does more than that. It actually will update your database with new card details if they are present, which is more than what the method name implies. I just wonder if that's surprising behavior based on the method name.

Francismori7:
I had the same hesitation but I figured that the method refers as much as possible to the Stripe's customer instance.

taylorotwell:
I think you should break this into two methods: `hasCardOnFile` and `syncCardWithStripe`.

Francismori7:
Works with me. Should we let the user call those methods by themselves as necessary in their specific applications?

Francismori7:
But `syncCardWithStripe` feels like this is a two-way sync, which it isn't. It just pulls from Stripe.

taylorotwell:
Sure, feel free to adjust method names as needed, but it should be two methods.

mlantz:
I agree two methods makes more sense a `has` anything should not update

Francismori7:
@taylorotwell This is what I came up with:

* `public function hasCardOnFile() : bool` **Determines if the user, according to the local DB, has a card linked.**
* `public function syncCustomerCardDetails : $this` **Forces the refresh of the customer's card details from Stripe's API.**
* `protected function fillCardDetails : $this` **Replaces the data about the card in the database for the Billable model.**

taylorotwell:
What about ->updateCardDetailsFromStripe()?

Francismori7:
Instead of the sync one? Sure even sounds better.

mlantz:
Agreed - I like the updateCardDetails is FromStripe needed?

Francismori7:
@mlantz IMO yes because updateCardDetails implies you can pass your own card details while this method will actually go through the API to retrieve data.

mlantz:
Oh yeah that makes total sense, agreed

Francismori7:
hasCardLocally is overkill, most features all rely on the local DB so we'd need to change them all. PHPDocs state that its local though

src/Billable.php
```diff
@@ -69,6 +69,61 @@ public function refund($charge, array $options = [])
...
+
+    /**
+     * Fills the user's properties with the source from Stripe.
+     *
+     * @param \Stripe\Card|null  $card
+     * @return $this
+     */
+    protected function fillCardDetails($card)
+    {
+        $this->card_brand = null;
```
taylorotwell: This looks like a possible change in behavior from the `updateCard` method, which did not `null` out the columns. Could be a breaking change of have unexpected behavior.

taylorotwell:
Cleaned up a few things and merged.

PR #299, opened 2016-05-30
Dates improved

Francismori7:
Invoice items now use Carbon for start and end dates in subscriptions (allows formatting and localization)
Invoice date now uses UTC for instantiation since Stripe's API uses UTC for all dates (see https://support.stripe.com/questions/what-timezone-does-the-dashboard-and-api-use)

### DIFF
--- a/resources/views/receipt.blade.php
+++ b/resources/views/receipt.blade.php
@@ -137,7 +137,10 @@
                     @foreach ($invoice->subscriptions() as $subscription)
                         <tr>
                             <td>Subscription ({{ $subscription->quantity }})</td>
-                            <td>{{ $subscription->startDate() }} - {{ $subscription->endDate() }}</td>
+                            <td>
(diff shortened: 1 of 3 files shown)


taylorotwell:
This appears to be a breaking change?

Francismori7:
Only the templates will change from the initial format to the default to string format of carbon. We may be able to set the default output format on the carbon instance it creates?

taylorotwell:
It's a breaking change if someone is calling $invoiceItem->startDate()... it was returning a string now its returning a Carbon instance.

Francismori7:
Right but in templates and used as a string, __toString is used, when using as a date, carbon will still work since it inherits from DateTime.

Or am I seeing this wrong?

Francismori7:
@taylorotwell Reworked my changes into two new methods instead. Non breaking changes now.

PR #303, opened 2016-06-08
[Feature] Multiple cards

Francismori7:
This requires a change in migrations, as seen in CashierTest.

Looking for some pointers as to how this could be implemented better.


taylorotwell:
Going to hold off on this for now.

peeyush1234:
Hi,

I am wondering if you got a chance to look into supporting multiple credit cards. I have a product requirement for supporting multiple cards in my project. I am unable to use Cashier because of this limitations. I am more than happy to help with this change if you would like. I can also create a new PR if this one is too old.

Best,
Peeyush

PS: Cashier is a great product

PR #304, opened 2016-06-08
Use Stripe's trial_end

### LINKED ISSUE #256: trial_ends_at not getting set
The snippet of code below in SubscriptionBuilder.php is returning NULL every time I create a subscription for a user. I don't believe $this->trialDays is every getting set. 

```
 'trial_ends_at' => $this->trialDays ? Carbon::now()->addDays($this->trialDays) : null,
```

Francismori7:
Make sure we get the right trial_ends_at data from Stripe itself. Stripe plans can have default trial periods, which will fix this issue. ->trialDays() will also still work since we get the end date from the subscription itself.

Fixes #256

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -168,7 +168,8 @@ class SubscriptionBuilder
         if ($this->skipTrial) {
             $trialEndsAt = null;
         } else {
-            $trialEndsAt = $this->trialDays ? Carbon::now()->addDays($this->trialDays) : null;
+            $trialEndsAt = $subscription->trial_end ? Carbon::createFromTimestampUTC($subscription->trial_end)
+                ->toDateTimeString() : null;
         }
 
         return $this->user->subscriptions()->create([


taylorotwell:
I don't really want to make this change in behavior. You should use Cashier's trial days method to override any trial days on the Stripe plan.

Francismori7:
This actually accounts for this. Cashier will send the trial data, if provided, and Stripe will return the subscription instance with the correct trial_end date, which we use in both cases to store our own trial_ends_at.

If trialDays was used, the payload sent to Stripe will override the default plan data.

This change only affects the stored data on our database. Let's say your Stripe plan has a 7 days trial default, and you create a subscription without calling trialDays(7), Cashier right now will set trial_ends_at to null, while it should set it to be in 7 days. This PR fixes this by using the subscription instance.

jrgilman:
@taylorotwell Does it really hurt to implement this? As @Francismori7 said it will get overridden by trialDays() if it was used. Otherwise it lets Stripe take the wheel.

chrisvanpatten:
> @taylorotwell wrote:
> You should use Cashier's trial days method to override any trial days on the Stripe plan.

It seems like this would still be the case. In fact, it's not about overriding at all — it's about using the default value from Stripe, if one is available. If you use Cashier out of the box, setting a trial length in Stripe is a pointless exercise since Cashier ignores it anyway.

Francismori7:
@chrisvanpatten is right. This is just about allowing the use of the defaults set through the Stripe dashboard. A developer will still be able to override or pass in a trial days period from the API which actually mimics the current behavior. 

Right now, not sending a trial period with the payload while your Stripe plan has a trial period set (dashboard set), cashier will save NULL in our database meaning the subscription would go on forever, however Stripe will CANCEL the subscription after the default period provided in the Stripe dashboard.

This PR just allows to really set the real end date returned by Stripe itself so they are perfectly in sync.

blazeworx:
👍 +1
I am also running into this problem. I have trial period set on the plan itself in the stripe dashboard. Cashier ignores this and sets trial_ends_at to NULL in the DB.

chrisvanpatten:
It's still bizarre to me that Cashier ignores the default value from Stripe, and I firmly believe this should be changed, but here's how I'm fixing it for the moment, in case anyone else wants this behaviour. This is in my billing setup controller. Theoretically you could 

```php
$plan_id = 'test-11oct16-level1';

// Retrieve the Stripe plan so we can fetch the trial length
$plan = \Stripe\Plan::retrieve($plan_id, ['api_key' => $user->getStripeKey()]);

// Create a new subscription
$user->newSubscription('primary', $plan->id)
        ->withCoupon(trim($request->input('coupon_value', '')))
        ->trialDays($plan->trial_period_days)
        ->create($request->input('stripe_token'), [ /* you get the idea */ ]);
```

PR #307, opened 2016-06-13
Use the config value when retrieving the Stripe key

aaronhuisinga:
In the Cashier documentation, it's mention that a requirement for getting setup for Stripe is to set the model and secret values in services.php. 

Currently when retrieving the Stripe key, Cashier is defaulting to using the getenv() function to retrieve the key instead of the config() function. 

When using the Laravel config cache option:
`php artisan config:cache`
Cashier stops working as the getStripeKey() function returns `false`.

This patch simply uses the config() function in place of the getenv() function, and fixes Cashier for everyone caching their config via Laravel. The solution was mentioned in #243

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -538,7 +538,7 @@ trait Billable
      */
     public static function getStripeKey()
     {
-        return static::$stripeKey ?: getenv('STRIPE_SECRET');
+        return static::$stripeKey ?: config('services.stripe.secret');
     }
 
     /**


Francismori7:
This will break the tests. Tests use the env variables to function properly.

mlantz:
Its also beside the point that the config itself just pulls the env anyway

aaronhuisinga:
@mlantz While this is true, the problem here is that Laravel offers the option to cache the config files via an Artisan command. When these are cached, the genenv() function returns null, and thus Cashier stops working. It seems strange to me that a supported Laravel function out of the box would break an extension written with Laravel compatibility in mind. I'll just leave my workaround code in my User model for now.

taylorotwell:
@aaronhuisinga makes a good point.

taylorotwell:
However, to keep from breaking BC we should probably check env first then fall back to config.

Francismori7:
PHP7 would've made this easier with `??`

Le jeu. 16 juin 2016 09:45, Taylor Otwell <notifications@github.com> a
écrit :

PR #309, opened 2016-06-22
Return description for single charges

socieboy:
I made a application using cashier, no using subscriptions but I do single charges for now.

When I generate the invoice I found there is not any methods or function to return the description of a single charge.
So I'm doing this PR and probably it's not the best way to handle that issue, but with the help of everyone we can make something great!

I'm glad to make my first PR to Laravel packages.

### DIFF
--- a/src/Invoice.php
+++ b/src/Invoice.php
@@ -314,8 +314,8 @@ class Invoice
     }
     
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can write your own helper to do this if needed.

PR #312, opened 2016-07-01
include tax percentage when swapping plans

### LINKED ISSUE #236: Applying subscription taxes when swapping?
You have made it very easy to create subscription with a `tax_percent` value, but I have found that this doesn't work in the context of swapping. Unless I'm mistaken. How would you go about doing this? Appreciated.

firsara:
and include tax percentage when creating invoices through Billable in general, fixes #236

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -211,7 +211,13 @@ trait Billable
     {
         if ($this->stripe_id) {
             try {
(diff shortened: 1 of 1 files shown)


taylorotwell:
This was intentional because different types of one off charges could have different tax rates. You should pass whatever tax rate you want in on each invoice.

firsara:
alright, thanks for pointing it out. I'm just currently overriding the function so no worries for me ;-)

Issue #320, opened 2016-08-10
[Proposal] 5.3 Tweak (param order in first())


BenGriffiths:
5.3 comes with a change to the ordering of parameters with the method `first()` on collections - this change is needed:

https://github.com/laravel/cashier/compare/6.0...BenGriffiths:BenGriffiths-patch-1

I've not made a pull request yet as I imagine this will need a new branch (i.e 6.1) as it breaks backward compatibility with Laravel 5.2

oyed:
Submit it as a PR. It's not really a proposal, it's a fix; Cashier in 5.3 just doesn't work because no `->subscription()` calls work anymore.

sagarrabadiya:
below code in Laravel\Cashier\Billable.php in subscription method should fix it.

```
return $this->subscriptions->sortByDesc(function ($value) {
            return $value->created_at->getTimestamp();
        })
        ->first(function ($key, $value) use ($subscription) {
            if($key instanceof Subscription) {
                return $key->name === $subscription;
            }
            return $value->name === $subscription;
        });
```

as its comparing if $key is instance of Subscription(laravel 5.3) else it will consider $value(laravel 5.2)

so it will be backward compitible as well.

BenGriffiths:
Closing this as it's not longer required and sorted in 7.0

PR #321, opened 2016-08-16
#320 fixed with backward compatibility to laravel 5.2

sagarrabadiya:
I have added a simple if to check the parameter order if its laravel 5.3 then first param will instance of Subscription otherwise second param will be instance of Subscription.


taylorotwell:
We will just have a new branch on release.

sagarrabadiya:
no worries 👍🏻

PR #325, opened 2016-08-20
removed extra else

sudiptpa:
Removed extra else after if, no else statement is required if (if) statement is not true, simply we can do:

` ``
       if (method_exists($this, $method)) {
            return $this->{$method}($payload);
        }
        
        return $this->missingMethod();
```

### CODE REVIEW

**Comment by taylorotwell** [comment]
I leave "extra" else's in place for visual indentation distinction.

**Comment by sudiptpa** [comment]
i think this looks more cleaner.

```php
        if (method_exists($this, $method)) {
            return $this->{$method}($payload);
        }
        
        return $this->missingMethod();
```

PR #336, opened 2016-10-04
Now using correct tax percentage

NiclasHedam:
A new invoice is now created with the right tax percentage as stated in the Billables taxPercentage()


taylorotwell:
This wouldn't be desired in some cases where you want to charge multiple tax rates for different items on an invoice.

Issue #334, opened 2016-09-27
[Proposal] Enabling a dynamic way of loading the Stripe model


Repox:
As another user also mentioned in  #332 I also need to use another model as the Stripe model.

Cashier currently doesn't support that, as the foreign key is hardcoded, but it's easily fixable.

This feature would allow for a flexible multi tenant architecture with Cashier.

PR #337, opened 2016-10-05
Enabled dynamic loading of Stripe model

Repox:
This is a pull request related to proposal #334 

I know that guidelines dictate that a proposal needs to be approved, but I think this PR will better explain the goal of the proposal.

One of the main changes other than the dynamic loading of the correct model and foreign key based on the configuration, I renamed a variable related to the model from `user` to `owner` (in lack of a better word).

For backward compatibility, the `Subscription::user` method remains, but utilizes `Subscription::owner`.

Also, updated the README to better match test setup conditions.

### DIFF
--- a/src/Invoice.php
+++ b/src/Invoice.php
@@ -234,7 +234,7 @@ class Invoice
     public function view(array $data)
     {
         return View::make('cashier::receipt', array_merge(
-            $data, ['invoice' => $this, 'owner' => $this->owner]
+            $data, ['invoice' => $this, 'owner' => $this->owner, 'user' => $this->owner]
         ));
     }
 
--- a/src/Invoice.php
+++ b/src/Invoice.php
@@ -11,11 +11,11 @@ use Symfony\Component\HttpFoundation\Response;
 class Invoice
 {
     /**
-     * The user instance.
+     * The Stripe model instance.
      *
      * @var \Illuminate\Database\Eloquent\Model
      */
-    protected $user;
+    protected $owner;
 
     /**
      * The Stripe invoice instance.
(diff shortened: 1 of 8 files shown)


src/Subscription.php
```diff
@@ -41,13 +42,22 @@ class Subscription extends Model
...
+        return $this->owner();
+    }
+
+    /**
+     * Get the model related to the subscription.
+     */
+    public function owner()
+    {
         $model = getenv('STRIPE_MODEL') ?: config('services.stripe.model', 'User');
+        $foreignKey = Str::snake(class_basename($model)).'_id';
```
taylorotwell: Why not create the model instance here and use `getForeignKey()`?

src/Invoice.php
```diff
@@ -234,7 +234,7 @@ protected function formatAmount($amount)
     public function view(array $data)
     {
         return View::make('cashier::receipt', array_merge(
-            $data, ['invoice' => $this, 'user' => $this->user]
+            $data, ['invoice' => $this, 'owner' => $this->owner]
```
taylorotwell: You'll need to still send `user` to the view as well to not break backwards compatibility.

taylorotwell: You should send both `owner` and `user` even though they are the same variable.

taylorotwell:
Any thoughts on feedback?

src/Invoice.php
Repox: You're right. In case of manual changes to the invoice, developers still referring to the `user` variable will have issues. Fixed that.

src/Subscription.php
Repox: This would of course be safer. I made a commit to fix that.

PR #338, opened 2016-10-10
Set Subscription Prorate Argument when using updateQuantity

howlowck:
Currently
`$subscription->noProrate()->updateQuantity($amount);` does not correctly set the prorate argument on the Stripe API when updating the subscription.

The proposed change automatically sets the `prorate` property that's on the Subscription model.


Francismori7:
Styling

PR #344, opened 2016-11-04
Add invoiceForUpcoming method

kitbs:
Add invoiceForUpcoming method to create invoice items without immediately paying them.
See proposal https://github.com/laravel/cashier/issues/335.

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -79,16 +79,17 @@ trait Billable
     }
 
     /**
-     * Invoice the customer for the given amount.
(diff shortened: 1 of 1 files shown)


taylorotwell:
Renamed it to `tab`, as in adding something to someone's bar tab.

dweb-x:
Why rename it? it's confusing and takes ages to find it. People are likely to type "invoice" and see what intellisense suggests. BTW, I found out about the feature from stackoverflow which uses the original name. I thought it was missing and was coming here to grab the code.

PR #356, opened 2016-11-25
Fix Skipping trial on swapping

themsaid:
As per the docs https://laravel.com/docs/5.3/billing#changing-plans

> If you would like to swap plans and cancel any trial period the user is currently on, you may use the  skipTrial method:

```
$user->subscription('main')
        ->skipTrial()
        ->swap('provider-plan-id');
```

However the Subscription model doesn't have a `skipTrial()` method, this PR adds the proper method.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -171,7 +171,7 @@ class Subscription extends Model
         $subscription = $this->asStripeSubscription();
 
         $subscription->quantity = $quantity;
-        
+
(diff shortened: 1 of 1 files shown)


dedycds:
It would be nice to make this available on Laravel Cashier ~6.0 for laravel 5.2, because in documentation it says that `skipTrial()` method exists when it actually not there.

PR #371, opened 2017-01-10
replaceCard() method rather than updateCard()

Mr-Anonymous:
May I suggest you to change the way the `updateCard()` method works or perhaps add a new method called `replaceCard()` that replaces the existing card?

**The problem with the current `updateCard` method:**

With the `updateCard()` method, the card is added to the sources list and then sets the new card as the `default_source`. the result of this method has 2 outcomes:

1. Multiple cards gets added to the list although the recent one is set as `default_source`

2. When updating the card using this method, if there are any unpaid invoices (i.e. invoices in `past_due` state), they are not getting automatically charged during the card update.

In order for stripe to re-attempt charging customer on all invoices in `past_due` state, the `source` parameter needs to be passed. So I have created a new method called `replaceCard()`. 

What do you think? Can this be added to Cashier?

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -401,6 +401,36 @@ trait Billable
     }
 
     /**
+     * Replace Customers Card by deleting the existing default source.
+     * and adding the new card as the source
+     *
+     * @return $this
+     */
+    public function replaceCard($token)
+    {
+        $customer = $this->asStripeCustomer();
+        $token = StripeToken::retrieve($token, ['api_key' => $this->getStripeKey()]);
+        // If the given token already has the card as their default source, we can just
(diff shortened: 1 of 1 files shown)


taylorotwell:
You are free to add this to your own billable class if needed.

Mr-Anonymous:
@taylorotwell can you tell me how I can extend the billable class without editing the `Billable.php` file directly so future updates doesnt affect it? Thanks.

taylorotwell:
Can’t you just define this method on your User object?

Mr-Anonymous:
I guess. At the moment, I have this method directly in my controller. No worries.. Thanks.

Issue #345, opened 2016-11-05
can't pass trial_end timestamp to Stripe


enkay:
I have a subscription that doesn't require a card upfront. As per the documentation, I set the trial_ends_at on the user model when they start their trial.

Now if the user upgrades before the end of the trial, I don't want Stripe to charge them until the end of the trial, so I want to set the **trial_end** parameter on the request.

Right now it seems the only way to set **trial_end**  is through the **trialDays()** method which is not very convenient as I already have the trial end timestamp, and I don't want cashier to calculate a new date based on the current time.

It seems there was a hidden method **trialFor()**  that could do this in previous versions but it was removed  and doesn't work when I try it.

kyranb:
Agree, this behaviour seems really off and has caused a lot of confusion from what I've seen on GitHub, The Forums, Reddit etc.

enkay:
I'm surprised this hasn't been brought up before as it seems like a very common subscription model for saas businesses: offer a free trial, and try to get users to upgrade to a paid subscription before their trial ends.

Is there any plans to better support this model? Or is there another workaround?

themsaid:
@enkay  can you please share some code samples? How do you instruct cashier to upgrade the user to a paid plan??

joshgriggs:
I have now also run into this problem. The context is as follows:

- A user signs up to our application without card information up front. The registration code manually creates their trial end by updating the trial_ends_at column in the database for that user. 
- When the user visits the billing page, they can see that they are on trial and can choose to add their card information to ensure that their subscription remains live after their trial end date
- When they submit their card information, the subscription is then created with the trialDays() method, with the number of days currently being calculated as the difference in days from their trial end date from now. This is problematic, because if the trial ends at date is 9 days 23 hours and 59 minutes away, it is calculated as being 9 days away thus effectively reducing the users' trial by a day.

It would be great if the trialFor() method could be re-implemented so that a timestamp / carbon date could be used instead of specifying days.

codegeek1001:
There is a skipTrial() function that lets you end a trial immediately which essentially nulls the trial_ends field.

enkay:
@codegeek1001 I don't want to cancel the trial, I want to start charging once their trial is over. I need a method to set the underlying Stripe parameter **trial_end** manually.

codegeek1001:
I understand what you are saying but in practice, when a client "upgrades", they are charged right away and it does not care how many trial days still remain. In other words, they become a paying client effective immediately. 

[...]

PR #377, opened 2017-02-08
Specify trial duration of a subscription with DateTime

mrsimonbennett:
Allows you to explicitly define when a trial will end.

#345


mrsimonbennett:
thanks @taylorotwell (y)

PR #382, opened 2017-02-19
Bump dompdf

barryvdh:
This version fixes the PHP 7.1 bug (numeric value encountered). See 
https://github.com/dompdf/dompdf/releases/tag/v0.8.0


barryvdh:
Note that I didn't test this, as I don't use Cashier. but it doesn't list any breaking changes.

PR #407, opened 2017-05-02
Stripe Trial End Date

dexter505:
The Stripe API will return an error if the trial end date is not in the future, for instance if you did:
```
$subscription->trialDays(0)->create();
//or
$subscription->trialUntil(Carbon::create(2010, 1, 31, 0))->create();
```
I added the function "setTrialExpires" which only sets $trialExpires if its a date in the future.

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -98,7 +98,7 @@ class SubscriptionBuilder
      */
     public function trialDays($trialDays)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Just let Stripe throw an error. The programmer should know.

dexter505:
The reason I made the change is before #377 trialDays(0) worked. 
If you're dynamically generating your # of trial days for a returning customer, it may not be apparent that you need to use skipTrial() instead. I think it would make cashier a little easier to use.

PR #420, opened 2017-06-30
Adding support for multiple subscriptions

joegarbett:
Adds support for multiple subscriptions (As documented @ https://stripe.com/docs/subscriptions/multiplan).

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -141,7 +141,7 @@ trait Billable
      */
     public function newSubscriptions($subscriptions)
     {
-        return new SubscriptionsBuilder($this, $subscriptions, "");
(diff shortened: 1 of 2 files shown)


taylorotwell:
We would need to think this through more.

joegarbett:
Okay but its a feature that's been available in Stripe for sometime now (~6 months) and isn't easily implementable currently.

roblesterjr04:
I second this. Multiple subscriptions can already be done with the package, but what I think Joe is asking for, and I am asking for, is adding multiple PLANS under a single subscription. Because I need to bill different rates under a single invoice. Seats/storage/usage etc.

PR #422, opened 2017-07-04
[Proposal] Merge w/ cashier-braintree & add gateway API

inxilpro:
**This PR is not meant to be merged. It's here to facilitate a discussion (using a PR allows for better visibility into proposed code changes).**

Right now there's not a huge difference between `laravel/cashier` and `laravel/cashier-braintree`. I took a day to see how hard it would be to merge the two. This PR isn't complete or tested, but I've made enough progress that I thought it would be worth discussing.

The basic idea is to refactor `stripe_id`/etc to `payment_gateway = "stripe"` and `payment_gateway_id`. Then, I've broken the Stripe- and Braintree-specific code into separate manager classes, and combined the shared code.

A backwards-compatibility layer is available right now, although I don't know if it's necessary (or if it should be released as a separate package).

Braintree- and Stripe-specific calls can be made via a `braintree()` or `stripe()` (or any other gateway) magic method. An `ifGateway()` method is also provided for packages that want to support multiple gateways, or if you're transitioning a code base from one to the other and want to support both for a period of time.

For example:

```php
$user->charge(1000); // Uses $user->payment_gateway or default gateway if none set

$user->stripe()->deletedCards(); // deletedCards currently Stripe-only

$user->ifGateway('stripe', function($stripe) {
  $stripe->deletedCards();
})->ifGateway('braintree', function($braintree) {
  $customer = $braintree->asGatewayCustomer();
  // Custom Braintree code to delete cards
});
```

## Basic overview

Each gateway implements the following:

  - `Gateway` — the base Gateway
  - `SubscriptionBuilder` creates new `Subscription` models
  - `SubscriptionManager` for things like `cancel()` and `resume()`
  - `BillingManager` for things like `charge()` and `refund()`

The `Billable` trait defers to the `BillingManager` for the model's configured gateway (or the default if none set) where appropriate. So where before `cashier` and `cashier-braintree` each implemented `invoiceFor()` differently in their `Billable` trait, now `Billable::invoiceFor()` calls `Billable::getBillingManager()->invoiceFor()`.

There's more to it, but that's the basic gist.

## A few things up for discussion:

  - Are you open to me continuing along this path? I think it can be done in a fairly non-breaking way, and would help open the door for other gateways/etc.
  - Does it make sense to break into 3 or 4 packages (cashier/cashier-stripe/cashier-braintree/cashier-compat)
  - My gut says that Cashier should exclusively use zero-decimal values for amounts (like Stripe) to allow for a more seamless transition between gateways. I've started to implement this is `Gateway::convertZeroDecimalValue` which lets each gateway convert a value into the format that it needs (so Braintree can convert `100` to `1.0`). This would be the biggest breaking change, but I think it makes a lot of sense.

Cheers!


inxilpro:
Wow. I just realized I branched from `7.0`. Hm. For sake of discussion I'm going to close and submit against that branch. Obviously this will need to be refactored to use the `master` code.

PR #423, opened 2017-07-04
[Proposal] Merge w/ cashier-braintree & add gateway API

inxilpro:
**This PR is not meant to be merged. It's here to facilitate a discussion (using a PR allows for better visibility into proposed code changes).**

Right now there's not a huge difference between `laravel/cashier` and `laravel/cashier-braintree`. I took a day to see how hard it would be to merge the two. This PR isn't complete or tested, but I've made enough progress that I thought it would be worth discussing.

The basic idea is to refactor `stripe_id`/etc to `payment_gateway = "stripe"` and `payment_gateway_id`. Then, I've broken the Stripe- and Braintree-specific code into separate manager classes, and combined the shared code.

A backwards-compatibility layer is available right now, although I don't know if it's necessary (or if it should be released as a separate package).

Braintree- and Stripe-specific calls can be made via a `braintree()` or `stripe()` (or any other gateway) magic method. An `ifGateway()` method is also provided for packages that want to support multiple gateways, or if you're transitioning a code base from one to the other and want to support both for a period of time.

For example:

```php
$user->charge(1000); // Uses $user->payment_gateway or default gateway if none set

$user->stripe()->deletedCards(); // deletedCards currently Stripe-only

$user->ifGateway('stripe', function($stripe) {
  $stripe->deletedCards();
})->ifGateway('braintree', function($braintree) {
  $customer = $braintree->asGatewayCustomer();
  // Custom Braintree code to delete cards
});
```

## Basic overview

Each gateway implements the following:

  - `Gateway` — the base Gateway
  - `SubscriptionBuilder` creates new `Subscription` models
  - `SubscriptionManager` for things like `cancel()` and `resume()`
  - `BillingManager` for things like `charge()` and `refund()`

The `Billable` trait defers to the `BillingManager` for the model's configured gateway (or the default if none set) where appropriate. So where before `cashier` and `cashier-braintree` each implemented `invoiceFor()` differently in their `Billable` trait, now `Billable::invoiceFor()` calls `Billable::getBillingManager()->invoiceFor()`.

There's more to it, but that's the basic gist.

## A few things up for discussion:

  - Are you open to me continuing along this path? I think it can be done in a fairly non-breaking way, and would help open the door for other gateways/etc.
  - Does it make sense to break into 3 or 4 packages (cashier/cashier-stripe/cashier-braintree/cashier-compat)
  - My gut says that Cashier should exclusively use zero-decimal values for amounts (like Stripe) to allow for a more seamless transition between gateways. I've started to implement this is `Gateway::convertZeroDecimalValue` which lets each gateway convert a value into the format that it needs (so Braintree can convert `100` to `1.0`). This would be the biggest breaking change, but I think it makes a lot of sense.

I realize now that I branched off of `7.0` rather than `master`. Doh. Obviously if I move forward I will refactor using the latest changes in `master`.

Cheers!


config/cashier.php
```diff
@@ -0,0 +1,18 @@
+<?php
```
inxilpro: Not sure if a config file will be needed. I added one early when I thought it'd be necessary.

database/2017_07_03_091748_upgrade_cashier.php
```diff
@@ -0,0 +1,31 @@
...
+	 * @return void
+	 */
+	public function up()
+	{
+	    // payment_gateway
+        // payment_gateway_id
+		// paypal_email
+
+        // payment_gateway_plan
+	}
```
inxilpro: It should be possible to offer a widely compatible migration using something along the lines of:

```php
$className = Cashier::getModelName();
$tableName = (new $className())->getTable();
Schema::($tableName, function( // ...
```

src/BraintreeService.php
```diff
@@ -0,0 +1,20 @@
+<?php
```
inxilpro: This is for compatibility reasons, but I don't like it :)

src/Cashier.php
```diff
@@ -3,10 +3,25 @@
...
+use Illuminate\Support\Str;
+use Laravel\Cashier\Exception as CashierException;
+use Laravel\Cashier\Gateway\Gateway;
 
+/**
+ * Class Cashier
+ *
+ * @package Laravel\Cashier
+ */
 class Cashier
```
inxilpro: I'm not a huge fan of having an all-static `Cashier` object. I'll explore options—maybe a facade.

src/Gateway/Braintree/Invoice.php
```diff
@@ -0,0 +1,286 @@
...
+
+namespace Laravel\Cashier\Gateway\Braintree;
+
+use Braintree\Transaction as BraintreeTransaction;
+use Carbon\Carbon;
+use Dompdf\Dompdf;
+use Illuminate\Support\Facades\View;
+use Symfony\Component\HttpFoundation\Response;
+
+class Invoice extends \Laravel\Cashier\Gateway\Invoice
```
inxilpro: `Invoice` instances need to be thought out a bit more.

src/Gateway/Braintree/SubscriptionBuilder.php
```diff
@@ -0,0 +1,144 @@
...
+     * Create a new Braintree subscription.
+     *
+     * @param  string|null $token
+     * @param  array $subscriptionOptions
+     * @return \Laravel\Cashier\Subscription
+     * @throws \Laravel\Cashier\Gateway\Braintree\Exception
+     */
+    public function create($token = null, array $subscriptionOptions = [])
+    {
+        $this->setBraintreeCustomerFromArgs(func_get_args());
```
inxilpro: This is for compatibility with `laravel/cashier-braintree`'s API.

inxilpro: Rather than passing the customer data to `create` you now chain `->customer()->create()`.

src/Gateway/Card.php
```diff
@@ -1,8 +1,6 @@
 <?php
 
-namespace Laravel\Cashier;
-
-use Stripe\Card as StripeCard;
+namespace Laravel\Cashier\Gateway;
 
 class Card
```
inxilpro: It may be possible to bring some of the Stripe-only functionality, like card management, into Braintree.

taylorotwell:
This will be too large for me to review anytime soon to be honest.

inxilpro:
@taylorotwell That's fair. Is there a better time to reassess?

omar10594:
This could be very useful for projects that use a different provider than stripe or braintree (or if they need to user more than only one provider).

Like in our case, we need to use Conekta for subscriptions on Mexico, but Stripe for subscriptions on USA.

For stripe we use Cashier, but, for Conekta we should use their api.

inxilpro:
@omar10594 I'll probably revisit this in a few months. I think it's quite possible and would be mostly backwards-compatible.

tomhatzer:
@inxilpro any news on this for the new cashier/spark version? :)

inxilpro:
We’re about to start refactoring some of our billing, so I’ll probably revisit soon. Not sure if it will be in the form of a cashier PR or not, though. @taylorotwell — is this something you’re open to, or is it too big of a change for cashier?

pascualstromsnes:
@inxilpro Hello there, I have just read through your proposal after realizing that adding support for a custom payment provider i Spark/Cashier wasn't as straight-forward as I had imagined.
I really love the idea of Cashier transcending into a more generic package with the possibility of developing any payment provider implementation one would want.

If this were to became a reality I think Spark would instantly become way more interesting for the European market at least. There are multiple payment providers here utilizing even cardless mobile payments that already have awesome API's and have also developed packages for popular OS software.

A more generic Laravel Cashier with support for any type of payment provider and also being available for use natively in Spark would be a dream come true! 💯

Issue #412, opened 2017-06-09
Changing plans does not change the local subscription name


TeRMiNaTe:
I am not sure what the use case is for a subscription to correspond to 2 different plans, but in my application when I want to upgrade a 'basic' subscription to a 'premium' one I would hope that the name would change as well, not only the plan identifier.  
The following leaves the user with a 'basic' subscription, but on a plan that is 2x more expensive
```
$this->user->subscription('basic')->swap('premium')
```
The quick solution I've found so far is to use this, rather inconveniently wherever I need the upgrade to happen
```
$this->user->subscription('basic')->swap('premium')->update(['name' => 'premium']);
```
Since altering the behaviour of swap is likely to be a breaking change for many application, can we get a different method that also renames the local subscription name for convenience when swapping?

olimorris:
I agree the default behaviour is odd.

Doesn't seem logical to update the ID in Stripe but not update the name.

The default behaviour renders `$this->user->subscribed('premium');` as false yet in Stripe, the user is subscribed to the Premium plan.

In the mean time it may be worth using putting a pull request in the Cashier documents to inform users of this.

MichaelHoughton:
I have added a $name parameter as a second parameter to the swap() method.  (See https://github.com/MichaelHoughton/cashier/commit/39661ad50cceffe0b19bf5d8552da12a2c4c944d)

Do you think this could help make this easier?  If so, I will submit a PR for this.

MichaelHoughton:
@TeRMiNaTe - the PR has been merged, you should be able to close this issue now.

TeRMiNaTe:
@MichaelHoughton Thanks.

tmoldri:
What about braintree version of cashier? This problem still exist there. Could we get an update there also? It is really annoying bug.

olimorris:
Has this fix been released in Cashier yet?

mkarnicki:
I haven't encountered this issue yet, but I'm using Braintree. I'm quite upset to see a half-ready solution (Cashier supports more than stripe) is accepted and the bug closed. If the issue is still there for Braintree, the issue should not be closed.

PR #446, opened 2017-10-04
Changing plans does not change the local subscription name (#412)

MichaelHoughton:
This PR fixes an issue as identified at issue #412.

It allows for a simple way to update the subscription name when swapping a plan.

If the PR is accepted, I will submit a separate PR to update the Laravel Cashier documentation.


driesvints:
It's best to revert this imo. This just doesn't makes any sense. Why would you want to change the name of a subscription during a plan swap? You can easily just change it at a later point if you really want to.

Issue #367, opened 2017-01-02
How to extend Subscription model?


Mr-Anonymous:
I have created couple of custom tables that has the `subscription_id` (ID from cashier subscription table) foreign key.  I would like to add the eloquent relation to my models and the same `hasMany` relation in Subscription model. How can I extend the subscription model from cashier please?

Remo:
I had to add a custom relation to the subscription model. Not sure where exactly you want to use this, but in my case it always starts with the user object. I was able to override the `subscriptions` method in the user class like shown below:

```php
public function subscriptions()
{
   return $this
      ->hasMany(YourSubscription::class, $this->getForeignKey())
      ->orderBy('created_at', 'desc');
}
```

MichaelHoughton:
@Mr-Anonymous - you can simply add a new Subscription model in your project, and extend the Cashier Subscription Model File.

e.g. make the following file in your app directory (or where ever your model files are stored):

```php
<?php

namespace App;

use Laravel\Cashier\Subscription as CashierSubscription;
use Illuminate\Database\Eloquent\Model;

class Subscription extends CashierSubscription
{
    public function yourCustomMethod()
    {
        // add them here
    }
}
```

You will then have all the functionality of the Laravel Subscription methods, plus any custom ones you need.

shanecp:
You also need to override the relationship on the User model.

```
	/**
	 * Get all of the subscriptions for the Stripe model.
	 *
	 * @return \Illuminate\Database\Eloquent\Collection
	 */
	public function subscriptions()
	{
		return $this->hasMany(\App\Subscription::class, $this->getForeignKey())->orderBy('created_at', 'desc');
	}
```

driesvints:
Yeah you should be able to achieve this by only overwriting the `subscriptions` method on the billable model.

paul-adams-985:
Superb. Perfect

JohnnyWalkerDesign:
@driesvints I understood the comments made by @MichaelHoughton and @shanecp, but I was confused by your comment. Could you clarify what you mean? Thank you.

driesvints:
@JohnnyWalkerDesign take a look at @shanecp's example above my comment. That way you can overwrite the subscription model. There's one more new place where you need to overwrite it though in the latest Cashier Stripe version: https://github.com/laravel/cashier-stripe/blob/12.x/src/SubscriptionItem.php#L43

But please remember we don't officially support this and it's at a risk of breaking changes.

[...]

PR #448, opened 2017-10-16
Allow for Subscription model extension

tomhughes-invento:
Extracted from https://github.com/laravel/cashier-braintree/pull/35 to just add the getter and setter for the subscription model, allowing it to be overridden in the constructor. (still an open issue in #367 ).

The original pull request on the braintree project seems to only have been declined because of issues around the other part of it (passing $options through updateCard), and the first part is something I'm having to do for my use case (extra relationships on the subscription model), so figured it's probably worth seeing if this would make it in on its own

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -20,6 +20,34 @@ use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 trait Billable
 {
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
Override the subscription method.

PR #449, opened 2017-10-16
Dompdf version fixed for php 7.1

carduz:
As said in https://github.com/dompdf/dompdf/issues/1272 the versions of dompdf before 0.8.1 do not work properly with php 7.1


carduz:
Could you release the update as new composer version?

carduz:
@taylorotwell Could you release the update as new composer version?

Issue #450, opened 2017-10-18
[Proposal] Introduce default_card utility method in Billable


andreiglingeanu:
That's something very often done at the application level because almost every app needs to display the [`last4`](https://stripe.com/docs/api/php#card_object) field of the used card to the user.

In my humble opinion, that's something worth having in the lib.

Thanks for considering it.

andreiglingeanu:
Got merged in #452.

PR #451, opened 2017-10-18
introduce default_card() and use it in updateCardFromStripe()

andreiglingeanu:
The follow up pull request of #450.

By the way, is there any convention for writing method names in Laravel community? I see both `snake_case` and `camelCase` being used allover the place.

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -364,6 +364,24 @@ trait Billable
     }
 
     /**
+     * Returns the default card of the Customer.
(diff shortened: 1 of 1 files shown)


andreiglingeanu:
Is there any reason why StyleCI doesn't like the guard `return null` at the end?

taylorotwell:
Doesn't follow formatting of entire Laravel ecosystem.

PR #455, opened 2017-10-20
Get the account balance/credit for the entity

ehel:
https://github.com/laravel/cashier/issues/454

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -634,9 +634,9 @@ trait Billable
         $balance = $this->asStripeCustomer()->account_balance;
 
         if ($balance) {
-            return Cashier::formatAmount(- $balance);
+            return Cashier::formatAmount(-$balance);
         }
 
-        return null;
+        return 0;
     }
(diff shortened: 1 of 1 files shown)


src/Billable.php
```diff
@@ -623,4 +623,20 @@ public static function setStripeKey($key)
...
+     */
+    public function credit()
+    {
+        $balance = $this->asStripeCustomer()->account_balance;
+
+        if ($balance) {
+            return Cashier::formatAmount(- $balance);
+        }
+
+        return null;
```
MichaelHoughton: I think this should return 0 instead of null if there is no balance.

ehel: https://github.com/laravel/cashier/pull/455/commits/e38a3e81d1f9d37d8e294b61ddb1c8a9b8abca3f

taylorotwell:
No plans on adding this to core. You can make the Stripe calls necessary yourself to get this information.

PR #468, opened 2017-11-02
[7.0] Updated to PHPUnit 6

gabriel-caruso:
PS: [these StyleCI](https://styleci.io/analyses/8Pgp4o) aren't related to this PR, so I won't apply them.


taylorotwell:
Tests fail.

gabriel-caruso:
@taylorotwell The tests only failed because of PHPUnit 6.0 [requires PHP 7.0](https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L25). Should we drop support to `5.6`?

Issue #479, opened 2018-01-01
[Proposal] Implementing `@subscribed('plan')` directive


yamenarahman:
I was thinking about implementing custom blade directives like 
```
@subscribed('main')
// only subscribers
@endsubscribed
```

Should that be fine ?

driesvints:
Yeah I can see this being useful. Feel free to retry your PR with the feedback from Taylor.

PR #480, opened 2018-01-01
Added @subscribed blade directive

yamenarahman:
#479

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -104,7 +104,8 @@ trait Billable
         ], $options);
 
         return StripeInvoiceItem::create(
-            $options, ['api_key' => $this->getStripeKey()]
+            $options,
+            ['api_key' => $this->getStripeKey()]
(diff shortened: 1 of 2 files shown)


taylorotwell:
Things like this would typically go in the `boot` method.

yamenarahman:
@taylorotwell That's what I did, I just thought to make it go in the `CachierServiceProvider` and open the door for any other blade extensions there.

taylorotwell:
It's not in the boot method.

yamenarahman:
The changes on the `Billable` trait are auto PSR2 fixes.

yamenarahman:
If this PR will be merged should I update the docs ?

taylorotwell:
This has some problems. First, `plan` is not the first argument of subscribed, secondly it would need to match the subscribed signature entirely so that it provides the same functionality.

PR #481, opened 2018-01-09
Add tax to subtotal

itsDamien:
Tax must be added to subtotal to have a correct discount value:

## Example

**Without the fix:**
Subtotal = **100$**
Coupon = **50$**
Tax = **10$**
Total = **60$**
`function discount()` = **40$** ⚠️ bad result

**With the fix:**
Subtotal = **100$**
Coupon = **50$**
Tax = **10$**
Total = **60$**
`function discount()` = **50$** 👍 good result

### DIFF
--- a/src/Invoice.php
+++ b/src/Invoice.php
@@ -120,7 +120,7 @@ class Invoice
      */
     public function discount()
     {
-        return $this->formatAmount($this->invoice->subtotal - $this->invoice->total);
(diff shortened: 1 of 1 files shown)


itsDamien:
@tsummerer Personally, I have outsourced the cashier source code to fit my project. It’s not so complicated and now I can do what I want.

PR #484, opened 2018-01-11
Add extra parameters to payload

itsDamien:
Sometime the payload need extra parameters to allow these kind of Stripe Subscription arguments:

`billing=send_invoice`
`application_fee_percent=2`
`source=src_123`


taylorotwell:
If this is called the twice the parameters are still hanging around from the previous request. Also, I would call it something like withOptions or something.

PR #491, opened 2018-01-26
Add defaultPayload argument to buildPayload

tomhughes-invento:
Adding this array to buildPayload allows for an out-of-the-box way of using things that Stripe accepts when creating a subscriber without having to specifically change Cashier every time - e.g. application_fee_percent - with no observed impact on any existing functionality around the subscription builder


taylorotwell:
To me it would be nicer to have a ->withOptions builder method or something.

PR #501, opened 2018-03-14
Fix anchorBillingCycleOn()

josiasmontag:
Cashier tries to set `billingCycleAnchor` on the Stripe Subscription object. This makes Stripe throw an exception: `Received unknown parameter: billingCycleAnchor`. The correct parameter name is `billing_cycle_anchor`. 

Also see: https://stripe.com/docs/api#update_subscription-billing_cycle_anchor

This bug was introduced by [this commit](https://github.com/laravel/cashier/commit/6fdbd62103a2c7bd3f346a76a2c82387aaba76b2#diff-dde586b4ea0798ea4a467b9eba070a7e).

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -243,7 +243,7 @@ class Subscription extends Model
         $subscription->prorate = $this->prorate;
 
         if (! is_null($this->billingCycleAnchor)) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks, Tagged 7.1.0.

PR #529, opened 2018-06-04
Add ability to anchor a new subscription to a billing cycle date.

gcw07:
This allows you to anchor a new subscription to a specific date. For example, this allows you to anchor a new subscription to the 1st day of the next month. This makes the renewal date of the customer become the 1st of the month instead of on the day they initially signed up.

This feature already exists within Cashier for swapping subscriptions, but it wasn't there for new subscriptions. This fixes that.

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -3,6 +3,7 @@
 namespace Laravel\Cashier;
 
 use Carbon\Carbon;
+use DateTimeInterface;
 
 class SubscriptionBuilder
 {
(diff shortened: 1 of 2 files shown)


taylorotwell:
But, how is it actually saved? What if I have an existing customer and I want to change their billing cycle? What would be the exact Cashier syntax?

gcw07:
This is intended for new customers only. Existing customers subscription dates would not be changed. For new customers, the syntax would be:

```
$user = User::create([
    'email' => 'taylor@laravel.com',
    'name' => 'Taylor Otwell',
]);

// Create Subscription
$user->newSubscription('main', 'monthly-10-1')
            ->anchorBillingCycleOn(new \DateTime('first day of next month'))
            ->create($this->getTestToken());
```

Now if you had an existing customer and that customer is swapping plans, then you could use the `anchorBillingCycleOn` method that already exists within Cashier's Subscription class.

PR #536, opened 2018-06-26
Default to show coupon name instead of code

nhtahoe:
When creating a coupon in Stripe the help text next to Coupon Id is: 

>  We recommend leaving this blank so we can generate an ID for you.

The generated codes are complex which seems to imply Stripe thinks you should make coupon codes hard to guess.

Showing coupon id on invoices by default instead of the more anonymous name seems like a good idea.

### DIFF
--- a/resources/views/receipt.blade.php
+++ b/resources/views/receipt.blade.php
@@ -152,9 +152,9 @@
                     @if ($invoice->hasDiscount())
                         <tr>
                             @if ($invoice->discountIsPercentage())
-                                <td>{{ $invoice->coupon() }} ({{ $invoice->percentOff() }}% Off)</td>
+                                <td>{{ $invoice->couponName() }} ({{ $invoice->percentOff() }}% Off)</td>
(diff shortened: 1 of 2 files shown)


taylorotwell:
What is the "name" property set to if you don't provide a name for the coupon?

nhtahoe:
Good question. I just attempted to create a coupon in the Stripe interface, and if you leave the name field blank, you get "A value is required" and cannot create the coupon:
download

taylorotwell:
I'm just wondering about older coupons created before that was a requirement. It did not used to be required.

nhtahoe:
Just checked one I created in 2016 and it shows up as "Not Provided" in the Stripe interface. So I see how that could be problematic. Perhaps a fallback to ID if there is no name set?

taylorotwell:
Yes, that is what I would suggest.

taylorotwell:
Re-open when you are ready to make that change.

PR #552, opened 2018-08-26
[Proposal] New collection to allow shortcuts to filters

jimmypuckett:
It would be nice to have shorthands to filter the subscriptions like...

`Laravel\Cashier\Subscription::all()->onTrial()`

instead of...

`Laravel\Cashier\Subscription::all()->filter->onTrial()`

I was not sure if you would see the value or just prefer to leave it out, so I am making this a `WIP` to see what you think before adding the other helpers & doing test.

If you like it, just let me know & I will polish it off.  Otherwise, please close.

Thanks!

### DIFF
--- a/src/SubscribersCollection.php
+++ b/src/SubscribersCollection.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Laravel\Cashier;
+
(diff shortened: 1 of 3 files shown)


taylorotwell:
No plans on adding this right now.

PR #561, opened 2018-10-05
[7.0] Handle more webhooks to keep database consistency

pierrocknroll:
Hi,

Since there are some data on the Cashier database, like "trial_ends_at"/"quantity"/"ends_at"/"card_last_four", ... , if something is done on the Stripe dashboard, the database must be updated to keep consistency.

This PR will handle some more webhooks than the customer.subscription.deleted provided by Cashier :

- **customer.deleted** : when a customer is deleted on the Stripe dashboard, all his subscriptions must be cancelled in Cashier and his card informations + his stripe_id must be emptied.

- **customer.subscription.updated** : when the quantity, the plan, the trial_end are updated on the Stripe dashboard, the database must be updated too.

- **customer.updated** : if the default card of a user is updated on the Stripe Dashboard, the card_brand & card_last_four have to be updated.

- **customer.source.deleted**: if the card of a user is deleted on the Stripe Dashboard, the card_brand & card_last_four have to be emptied.

Let me know if there are other webhooks to handle or if I misunderstood something !

Thanks

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -175,7 +175,7 @@ class WebhookController extends Controller
     }
 
     /**
-     * Handle customer source deleted
+     * Handle customer source deleted.
      *
      * @param  array $payload
      * @return \Symfony\Component\HttpFoundation\Response
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -115,7 +115,7 @@ class WebhookController extends Controller
                 if (isset($data['trial_end'])) {
                     $trial_ends = Carbon::createFromTimestamp($data['trial_end']);
 
-                    if ($subscription->trial_ends_at->ne($trial_ends)) {
+                    if (! $subscription->trial_ends_at || $subscription->trial_ends_at->ne($trial_ends)) {
                         $subscription->trial_ends_at = $trial_ends;
                     }
                 }
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -5,9 +5,9 @@ namespace Laravel\Cashier\Http\Controllers;
 use Exception;
 use Illuminate\Http\Request;
 use Laravel\Cashier\Cashier;
+use Illuminate\Support\Carbon;
 use Stripe\Event as StripeEvent;
 use Illuminate\Routing\Controller;
-use Illuminate\Support\Carbon;
 use Symfony\Component\HttpFoundation\Response;
 
 class WebhookController extends Controller
@@ -57,7 +57,7 @@ class WebhookController extends Controller
     }
 
     /**
-     * Handle deleted customer
+     * Handle deleted customer.
      *
      * @param  array $payload
      * @return \Symfony\Component\HttpFoundation\Response
@@ -78,7 +78,7 @@ class WebhookController extends Controller
                 'card_brand'     => null,
                 'card_last_four' => null,
                 'trial_ends_at'  => null,
-                'stripe_id'      => null
+                'stripe_id'      => null,
             ])
                 ->save();
         }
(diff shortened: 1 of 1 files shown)


src/Http/Controllers/WebhookController.php
```diff
@@ -55,6 +56,125 @@ protected function handleCustomerSubscriptionDeleted(array $payload)
...
+                    ->markAsCancelled();
+            });
+
+            $user->forceFill([
+                'card_brand'     => null,
+                'card_last_four' => null,
+                'trial_ends_at'  => null,
+                'stripe_id'      => null,
+            ])
+                ->save();
```
stefanbauer: This line looks weird.

src/Http/Controllers/WebhookController.php
```diff
@@ -55,6 +56,125 @@ protected function handleCustomerSubscriptionDeleted(array $payload)
...
+    /**
+     * Handle customer updated.
+     *
+     * @param  array $payload
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    protected function handleCustomerUpdated(array $payload)
+    {
+        $user = $this->getUserByStripeId($payload['data']['object']['id']);
+        /* @var $user \App\User */
```
stefanbauer: This should go one line above.

src/Http/Controllers/WebhookController.php
```diff
@@ -55,6 +56,125 @@ protected function handleCustomerSubscriptionDeleted(array $payload)
...
+
+            // Change card details
+            if (isset(
+                $data['default_source'],
+                $payload['data']['previous_attributes']['default_source'],
+                $data['sources']['data']
+            )) {
+                $default_card = $data['default_source'];
+                foreach ($data['sources']['data'] as $card) {
+                    if (isset($card['id']) && $card['id'] == $default_card) {
```
stefanbauer: Maybe it's better here to flip the condition and do an "early continue"

driesvints:
FYI: I'm planning to thoroughly review this but probably won't get to it until next week.

In the meantime: can you perhaps send this to master instead? I'm afraid it might break people's setups on a minor release.

pierrocknroll:
I think it's on master now

src/Http/Controllers/WebhookController.php
```diff
@@ -55,6 +56,149 @@ protected function handleCustomerSubscriptionDeleted(array $payload)
...
+            )) {
+                $default_card = $data['default_source'];
+                foreach ($data['sources']['data'] as $card) {
+                    if (! isset($card['id']) || $card['id'] != $default_card) {
+                        continue;
+                    }
+                    $user->forceFill([
+                        'card_brand'     => $card['brand'] ?? null,
+                        'card_last_four' => $card['last4'] ?? null,
+                    ])->save();
```
driesvints: Isn't it best to use the `updateCardFromStripe` method here? I'm actually not sure myself if it's best to use that method or to rely on the data of the webhook.

src/Http/Controllers/WebhookController.php
```diff
@@ -55,6 +56,149 @@ protected function handleCustomerSubscriptionDeleted(array $payload)
...
+        /* @var $user \App\User */
+        $user = $this->getUserByStripeId($payload['data']['object']['customer']);
+
+        if ($user) {
+            $data = $payload['data']['object'];
+
+            if ($user->card_brand == $data['brand'] && $user->card_last_four == $data['last4']) {
+                $user->forceFill([
+                    'card_brand'     => null,
+                    'card_last_four' => null,
```
driesvints: Don't indent the arrows here.

src/Http/Controllers/WebhookController.php
pierrocknroll: I'm not sure too. I don't think it's very useful to make an extra call to Stripe here but maybe I'm wrong.

pierrocknroll:
Thanks @driesvints for your reviews.

And thanks @taylorotwell for the merge
I saw your changes, they are great !

Issue #476, opened 2017-12-14
only 10 invoice items shown


maxx1337:
If you have many invoice items, only the first 10 items will be shown on the invoice.

Stripe imitates the requests to 10 by default and offer pagination
https://stripe.com/docs/api#pagination

my workaround currently is this:
`if ($stripeInvoice->lines->has_more){`
`    $stripeInvoice->lines = $stripeInvoice->lines->all(['limit' => 100]);`
`}`
and pass then the $stripeInvoice to the Laravel\Cashier\Invoice constructor

rental-ninja:
Hi!
We are facing the same issue since many months ago. Some of our invoices have many lines and due to the fact that the findInvoice method only retrieves the invoice without all the lines, we are missing many lines in our invoices.
We tried to develop a solution without modifying neither the Cashier nor Spark, so we can keep our dependencies updated automatically. We could modify the invoice when a user is trying to download it, but the problem comes when we want to modify the notification (email) that Spark sends to the user when the subscription is modified (or just charged the normal subscription). We can't replace the notification nor the Controllers.
We will do a pull request to try to solve this issue as we need to get our invoices right for our customers. Thanks for the suggestion @maxx1337

driesvints:
PR is merged now so should be fixed in the latest release.

rental-ninja:
Great thanks a lot!

jimiero:
This is what I did, and seems to be working just fine:

`$items = auth()->user()->invoices();
        $total = $items->count(); // Specify your total here

        $perPage = 10;

        $currentPage = LengthAwarePaginator::resolveCurrentPage();
        
        $currentItems = $items->slice(($currentPage - 1) * $perPage, $perPage)->values();

        $invoices = new LengthAwarePaginator(
            $currentItems, // The items for the current page
            $total,        // Total number of items
            $perPage,      // Items per page
            $currentPage,  // Current page number
            ['path' => LengthAwarePaginator::resolveCurrentPath()] // Path to append to pagination links
        );

        $invoices->appends(request()->query());`

PR #562, opened 2018-10-05
[7.0] Bug solution for Issue #476

rental-ninja:
Now retrieve an invoice with ALL line items:
As per Stripe API documentation, when retrieving an invoice, you only retrieve "the first handful of those items", referring to the line items. Currently, the "findInvoice" method of the Billable only retrieves the invoice without retrieving all the lines. Therefore, when bills are generated, we get lots of missing lines and therefore it's impossible to check by our customers if their bill is OK.

This pull request solves this problem by simply retrieving up to 1000 lines of the invoice and replacing the lines object of the original invoice.

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -262,7 +262,10 @@ trait Billable
     public function findInvoice($id)
     {
         try {
-            return new Invoice($this, StripeInvoice::retrieve($id, $this->getStripeKey()));
+            $stripe_invoice = StripeInvoice::retrieve($id, $this->getStripeKey());
+            $stripe_lines = StripeInvoice::retrieve($id)->lines->all(array('limit'=>1000));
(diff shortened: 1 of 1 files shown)


rental-ninja:
Hi!
Thanks for your answer. No, definitely it's not needed 1000, it was only to make sure that we get all of them.
Our case is that we run with "seats". So, in Stripe we just have 2 plans and we multiply the amount of the plan per the number of "seats". So every change in this number creates two lines.
For the moment the maximum number of lines we had in an invoice was about 50. So I think that 100 would be OK but if this could be something we can adjust, for instance with a parameter, it would be nice.
Thanks for your time!

rental-ninja:
Great!
Thank you all for your help!

PR #563, opened 2018-10-09
[8.0] Revert changing subscription name on plan swap

driesvints:
This reverts https://github.com/laravel/cashier/pull/446

It doesn't really make sense to do this in the same operation. Swapping plans just means switching from one plan to the other. Changing the name of a subscription is just renaming it to a different context. Both have nothing to do with each other.

You can easily change the name of the subscription later on after the swap if you really want to. Changing this back keeps things more simple as well for swapping plans.

```php
$subscription = $subscription->swap(/* new plan */);

$subscription->name = 'new name';
$subscription->save();
```

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -232,10 +232,9 @@ class Subscription extends Model
      * Swap the subscription to a new Stripe plan.
      *
      * @param  string  $plan
-     * @param  string  $name
      * @return $this
      */
-    public function swap($plan, $name = null)
+    public function swap($plan)
     {
         $subscription = $this->asStripeSubscription();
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Couldn't they be using it for display purposes? Is it causing problems?

driesvints:
I just personally feel that this is doing too much in a single method. It just complicates the api with an unnecessary extra argument. This method is meant for swapping plans and not updating your subscription. I also don't see why swapping plans would change the "meaning/name" of a subscription. 

But feel free to close this if you still feel this could be useful. It would definitely not cause problems. I personally just wouldn't add a "name" argument as it isn't solving any real issues.

PR #570, opened 2018-10-11
[7.0] Bug solution: Invoice template suggestion of change

rental-ninja:
We have our own template for the invoices. However, regarding our last Pull request #564, we just realized that if you are using the template from the cashier (receipt.blade.php), when the number of lines exceeds the page size, there is a problem of visualization. This is because the entire "Invoice table" is within a single row of the general table of the document. So we just moved the entire invoice table outside the row.

We suggest this very simple change in the template for the users that are using the provided template for the invoice in the cashier. This is only a suggestion, as mentioned, we are using our own template. We have checked the result of this template and we like it because the Invoice table is much wider, taking profit of the whole page width. So now there is much more space for the "Description" column and for the "Date", wherein our case sometimes they were overlapped.

Be aware that this change will affect not only invoices with a lot of lines but also the rest of them. But apart from solving the problem for long invoices, we like this design, much spaced.

### DIFF
--- a/resources/views/receipt.blade.php
+++ b/resources/views/receipt.blade.php
@@ -112,72 +112,72 @@
                 @endif
 
                 <br><br>
-
-                <!-- Invoice Table -->
-                <table width="100%" class="table" border="0">
-                    <tr>
-                        <th align="left">Description</th>
-                        <th align="right">Date</th>
-                        <th align="right">Amount</th>
-                    </tr>
-
-                    <!-- Existing Balance -->
-                    <tr>
-                        <td>Starting Balance</td>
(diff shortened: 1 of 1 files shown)


rental-ninja:
Sure!
I'm trying to find a Formatting file for my IDE (IntelliJ) or an explanation of your formatting rules in your documentation but I can't find it. Could you please provide one of the two?
Thanks in advance!

driesvints:
We use StyleCI for checking our coding standards. It has a Laravel preset with all the fixers: https://docs.styleci.io/presets#laravel

taylorotwell:
Please re-submit with those fixes.

PR #579, opened 2018-11-12
[8.0] Update to v6 of the Stripe SDK

jshayes:
The [list of breaking changes](https://github.com/stripe/stripe-php/blob/master/CHANGELOG.md#600---2018-02-07) doesn't seem to impact this library.

The tests pass when I run them locally after this change, although it doesn't seem like the tests run on travis for this repo.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "illuminate/database": "~5.3",
         "illuminate/support": "~5.3",
         "nesbot/carbon": "~1.0",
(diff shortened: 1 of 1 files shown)


aaronhuisinga:
@taylorotwell It looks like v8.0.0 was tagged for release before this was merged in. Can we get a v8 release that updates to v6 of the Stripe SDK? That would be really great!

PR #581, opened 2018-11-18
Make sure to reactivate subscription when swapping plans

Francismori7:
When a user has a cancelled subscription, and within a UI, swaps his plan selection, we must make sure to reactivate the subscription, otherwise Cashier will take into account that the user has a "valid subscription", and the Stripe webhook that comes a few seconds later will immediately re-cancel the plan.

Adding `$subscription->cancel_at_period_end = false;` makes sure the plan is reactivated.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -253,7 +253,7 @@ class Subscription extends Model
         $subscription->plan = $plan;
 
         $subscription->prorate = $this->prorate;
-        
+
         $subscription->cancel_at_period_end = false;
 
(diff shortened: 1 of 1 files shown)


driesvints:
Hmm, I believe you're right. But this should be a bugfix targeted to 8.0, can you send it to there?

Can you also perhaps add a comment above this line so it's clear why we do this? Something in the nature of "If the plan was cancelled before we need to be sure to reactive the subscription..."

PR #582, opened 2018-11-20
[8.0] Make sure to reactivate subscription when swapping plans

Francismori7:
When a user has a cancelled subscription, and within a UI, swaps his plan selection, we must make sure to reactivate the subscription, otherwise, Cashier will take into account that the user has a "valid subscription", and the Stripe webhook that comes a few seconds later will immediately re-cancel the plan.

Adding `$subscription->cancel_at_period_end = false;` makes sure the plan is reactivated.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -254,6 +254,9 @@ class Subscription extends Model
 
         $subscription->prorate = $this->prorate;
 
+        // Let's make sure the subscription is resumed if the subscription was cancelled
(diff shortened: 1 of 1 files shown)


driesvints:
@Francismori7 can you link to the issue where we discussed this? I can't seem to find it anymore.

driesvints:
Nvm, found the PR. Anyways this looks ok.

PR #601, opened 2018-12-31
Correct PHP Doc @return tag

vpratfr:
$this->invoice returns a Stripe invoice object, and thus $invoiceFor too.


driesvints:
Thanks!

vpratfr:
You are welcome, but maybe you should consider wrapping those stripe invoice into cashier invoice objects directly?

driesvints:
@vpratfr if there's no specific reason for it I don't see why.

driesvints:
@vpratfr hmm actually on second hand it does might be better too. Might be something for a next release.

PR #603, opened 2019-01-02
[9.0] Test against latest Stripe API version

driesvints:
This makes sure we always use a fixed API version when making Stripe calls. It prevents breakages if another API version is defined in the Stripe account.


driesvints:
PR was changed to not enforce the API version but rather just test the latest API version in the test suite. This way people may still use a lower version if they need it for other Stripe related implementations.

PR #606, opened 2019-01-13
[9.0] Allow Carbon 2

kylekatarnls:
This will allow CarbonImmutable and all features added in Carbon 2 and so use the default version from Laravel 5.8.
Note that the carbon dependency could even be omitted as illuminate/support already requires it.


kylekatarnls:
Rebased to master #607

PR #607, opened 2019-01-13
[9.0] Allow Carbon 2

kylekatarnls:
This will allow CarbonImmutable and all features added in Carbon 2 and so use the default version from Laravel 5.8.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "dompdf/dompdf": "^0.8.0",
         "illuminate/database": "~5.7",
         "illuminate/support": "~5.7",
-        "nesbot/carbon": "~1.0",
+        "nesbot/carbon": "~1.0 || ~2.0",
         "stripe/stripe-php": "~6.0",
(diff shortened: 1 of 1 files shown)


driesvints:
@kylekatarnls if the package is directly used in the codebase (which it is) it needs to be an explicit dependency.

kylekatarnls:
OK, so first you could merge this to allow Carbon 2.

In a second move `Carbon::now()` could be replaced by `Date::now()` (and same for createFromTimestamp and createFromTimestampUTC to handle CarbonImmutable), `Carbon` typing could be replaced (idealy) by `DateTimeInterface` (then `Carbon::instance()` can be used to convert it).

In fact ideally, Laravel libraries that use Carbon should now rely on illuminate/support directly and get the date type from it. It seems faisable as there is only 3 different Carbon methods called. And in fact I should inspect but createFromTimestampUTC should probably be used everywhere instead of createFromTimestamp as it seems all timestamps are saved the same way and have no reason to not be UTC.

PR #612, opened 2019-02-27
[9.0] Do not clear trial_ends_at value to NULL

hafezdivandari:
Clearing an existing `trial_ends_at` value to NULL breaks the user's "generic" trial period.

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -159,7 +159,6 @@ class WebhookController extends Controller
 
             $user->forceFill([
                 'stripe_id' => null,
-                'trial_ends_at' => null,
                 'card_brand' => null,
                 'card_last_four' => null,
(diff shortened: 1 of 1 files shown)


driesvints:
I *think* this is indeed a valid change. I believe the trial_ends_at should indeed not be affected when a customer was deleted in Stripe.

taylorotwell:
No plans to make this change on a point release in case there are unexpected billing behavior changes for existing applications.

hafez-bsup:
@taylorotwell should I send this PR again for another branch? `master` maybe?

hafezdivandari:
@taylorotwell should I send this PR again for another branch? `master` maybe?

driesvints:
@hafezdivandari yes please do so.

hafezdivandari:
@driesvints Done: #614

PR #614, opened 2019-03-04
[10.0] Do not clear trial_ends_at value to NULL

hafezdivandari:
Clearing an existing `trial_ends_at` value to NULL breaks the user's "generic" trial period.


taylorotwell:
Can you describe your use case a little more and how the customer gets deleted from Stripe in your application?

PR #617, opened 2019-03-15
[9.0] Update invoice date method to work with latest '2019-03-14' stripe api.

mirago:
Hello, stripe have updated their api. The date method on the invoice fails, because they have renamed `date` field to `created`. 

I have amended the date method to use this `created` field.

### DIFF
--- a/src/Invoice.php
+++ b/src/Invoice.php
@@ -45,7 +45,7 @@ class Invoice
      */
     public function date($timezone = null)
     {
-        $carbon = Carbon::createFromTimestampUTC($this->invoice->date);
+        $carbon = Carbon::createFromTimestampUTC($this->invoice->created);
 
         return $timezone ? $carbon->setTimezone($timezone) : $carbon;
     }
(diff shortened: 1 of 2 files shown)


driesvints:
Is the BC with the previous api versions? Because otherwise all other apps from other users will break.

mirago:
Yes, I believe it is. I run the tests with version '2019-02-19' and see the screenshot. Both 'created' and 'date' are there.

Screenshot 2019-03-15 at 17 19 48

However when running tests with version 2019-03-14 the date field is no longer in the response.

I am not sure what process do you have with changes like this. Would you add the code to check if one of them exists, or rely on stripe?

taylorotwell:
@driesvints can you confirm this?

driesvints:
It's a bit unfortunate that Stripe doesn't shows docs on old api versions so I can't verify how long the `created` field has been around unless I actually manually test every version out. I've pushed a commit similar to yours that fixes this but with a `??` check so people on older versions still use the old `date` field. We might want to remove this in an upcoming version to drop support for older Stripe versions.

driesvints:
Fix will be released tomorrow btw.

PR #618, opened 2019-03-15
[9.0] Add support for options on subscription swap

### LINKED ISSUE #550: How can I add trial days to an existing subscription?
I want to add trial days after I create a subscription, so that the first payment is made when I create it and the next payment when the trial days are over.

Any ideas?

### LINKED ISSUE #580: Add ability to add coupon in swap
Currently it is not possible to add coupon while changing a plan using swap. We offer coupons to users and promote migrating to yearly plans.

keithbrink:
This resolves #580 - you can now pass in parameters when swapping a subscription, such as a coupon.

I believe this would also resolve #550, but I haven't tested it.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -37,7 +37,7 @@ You can set these variables in the `phpunit.xml` file.
 
 #### Coupons
 
-    * coupon-1 ($5)
+    * coupon-1 ($5) (Duration longer than once)
 
 ## Contributing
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
Please remove your comment.

keithbrink:
Done.

driesvints:
This is a breaking change on the `swap` method for people who are overriding this and is best to be sent to the master branch.

PR #623, opened 2019-03-19
Proposal for Stripe bootstrapping

### LINKED ISSUE #622: Bootstrap test environment
It would be beneficial if we could bootstrap the requirements to run the test suite which are detailed [in the `README.md`](https://github.com/laravel/cashier#running-cashiers-tests-locally). That way we wouldn't have to do everything manually.

There's also a problem with concurrency atm. where multiple builds running against the same test account could conflict each other. If we could dynamically set the IDs and counters for the tests we could concurrently run the test suite without every running into conflict.

This can be tested by running the test suite at the same time in separate processes.

keithbrink:
This is a proposal to resolve #622, adding bootstrapping to set up the proper Stripe test environment.

### DIFF
--- a/tests/CashierTest.php
+++ b/tests/CashierTest.php
@@ -67,17 +67,17 @@ class CashierTest extends TestCase
         ]);
 
         // Create Subscription
-        $user->newSubscription('main', 'monthly-10-1')->create($this->getTestToken());
+        $user->newSubscription('main', $this->plan_1_id)->create($this->getTestToken());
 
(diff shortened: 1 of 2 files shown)


driesvints:
Heya. Thanks for taking a stab at this. Unfortunately this doesn't helps with concurrency. You're still re-using items with the same ids as the one in other tests. To make this work we should work with generating random ids and setting static properties which we can re-use.

driesvints:
Btw, since this only affects the test suite you can send this to the 9.0 branch.

PR #624, opened 2019-03-19
Bootstrap stripe tests

keithbrink:
Another attempt at #622, this time supporting concurrency. I ran two tests at once without any problems.


driesvints:
@keithbrink I've sent in a follow-up PR which refactors this a bit towards static properties and setting up everything before the first test is run and removes items again after the last one has run: https://github.com/laravel/cashier/pull/625

Thanks a lot for your work on this!

PR #641, opened 2019-04-12
[9.0] Handle card failure in plan swap

driesvints:
Previously, when a plan swap was attempted and payment failed, the exception was cascaded to the end user and the update to the subscription in the app was not performed. However, the update to the subscription in Stripe itself was performed and the two states would be out of sync unless you implemented webhooks.

We've decided to catch the card failure exception and allow the plan swap to continue regardless of the failed payment. This leaves the subscription in a "past_due" state. This is because payment failure will be handled by Stripe and Stripe may attempt to retry the payment later on. When payment finally fails on its last attempt Stripe will send out a webhook to update the subscription in the way you specified in its settings: https://stripe.com/docs/billing/lifecycle#settings

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -10,6 +10,14 @@ To accommodate for this new behavior from now on Cashier will cancel that subscr
 
 If you were relying on catching the `\Stripe\Error\Card` exception before you should now rely on catching the `Laravel\Cashier\Exceptions\SubscriptionCreationFailed` exception instead. 
 
+### Card failure upon plan swapping
+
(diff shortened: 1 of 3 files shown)


hlorofos:
@driesvints what if a developer want to catch this specific event and perform its own business logic about failed plan update? Like in my case I want to revert back to previously used plan.

driesvints:
@hlorofos unfortunately we can't provide for two cases. You'll have to roll your own solution and override the method. We encourage making use of webhooks to let Stripe retry the payments first.

hlorofos:
Understandable and this is exactly how I did it. Just to explain this to anyone reading this thread, you have to create custom Billable trait and use your own Subscription model as the relationship. Then you can define any custom logic.
My reasonale is the following - a customer is on plan and trying to upgrade to a better plan, however the charge isn't successful (like there is no funds for such a costly plan). In my case I'm rolling back previous plan, so that customer could continue using it and potentially be re-billed with the usual price.

PR #668, opened 2019-05-13
Adding in an optional parameter to swap method so users can update metadata when changing plans

bkilshaw:
Currently the SubscriptionBuilder allows you to add metadata to a subscription with the ->withMetadata([]) method. However, there is no way to modify the subscriptions metadata while swapping plans.

This is beneficial as the metadata is often related to the subscription plan itself, so if a user needed to include metadata when creating the subscription there is a good chance they will need to update that metadata when they swap plans.

This is fully backwards compatible as the $metadata parameter defaults to a blank array, and setting $subscription->metadata = [] has no effect in Stripe, so it will not overwrite or delete existing metadata attributes, it only allows you to update or add new metadata if a value is passed in.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -348,11 +348,13 @@ class Subscription extends Model
      * @param  string  $plan
      * @return $this
      */
-    public function swap($plan)
+    public function swap($plan, $metadata = [])
     {
         $subscription = $this->asStripeSubscription();
 
         $subscription->plan = $plan;
+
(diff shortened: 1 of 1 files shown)


driesvints:
Unfortunately this is a breaking change for anyone overriding the `swap` method because you've added a new argument to the method signature. I get that this is beneficial but it's probably best that you send this in to master. Thanks.

PR #672, opened 2019-05-17
[10.0] Register webhook route with the package

driesvints:
This commit auto-registers the webhook endpoint with the package so the user doesn't has to do this themselves. They can still override it in their own routes file. Benefit of this is also that our integration test is now an actual endpoint test.

In an upcoming PR another route for payment intents will be added.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -1,0 +1,5 @@
+<?php
+
+use Illuminate\Support\Facades\Route;
+
+Route::post('webhook', 'WebhookController@handleWebhook')->name('webhook');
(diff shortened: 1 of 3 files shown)


taylorotwell:
I think you may want to do stuff that other providers that register routes do like check if routes are already cached.

driesvints:
@taylorotwell good spot. I actually went through the service provider a bit to clean it up. I've also implemented the routes now in the same way as Horizon. It now makes use of the `loadRoutesFrom` method.

PR #678, opened 2019-05-28
[9.0] Remove strict type check for trialUntil()

cmorbitzer:
This removes the type check for a `Carbon` object in `trialUntil()` because those using Carbon 2 may also pass a `CarbonImmutable` object.

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -39,7 +39,7 @@ class SubscriptionBuilder
     /**
      * The date and time the trial will expire.
      *
-     * @var \Carbon\Carbon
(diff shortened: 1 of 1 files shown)


src/SubscriptionBuilder.php
```diff
@@ -118,7 +118,7 @@ public function trialDays($trialDays)
      * @param  \Carbon\Carbon  $trialUntil
```
driesvints: We'll also need to change the method signature here and on the property above in the class like so:

```suggestion
     * @param  \Carbon\Carbon|\Carbon\CarbonImmutable  $trialUntil
```

cmorbitzer: Done! I used `CarbonInterface` instead as it's more future-proof.

PR #685, opened 2019-06-07
[10.0] Refactor Invoices

driesvints:
These changes utilize the Money library to format money values from the invoice and invoice line objects. It makes use of the invoice's currency to make sure there's no currency conflict when trying to display the values.

New Composer requirements:

- moneyphp/money: used for formatting money values
- symfony/intl: used for the above library to properly format money values
- (optional) ext-intl: used for more locales when formatting money values

This brings along some changes as well:

- Removed the $currencySymbol setting on the Cashier object along with the useCurrencySymbol and usesCurrencySymbol methods
- Removed the $symbol parameter from the useCurrency method on the Cashier object and its guessCurrencySymbol method
- Refactored the formatAmount method on the Cashier object to accept an optional $currency parameter. By default it'll use the current set Currency. Also refactored its internals to use the Money library to format the value.
- The starting balance is now no longer subtracted from the subtotal of an invoice
- The rawTotal method now returns an integer instead of a float
- A new tax() method is added to the Invoice object which returns the invoice tax formatted with its currency

And finally the invoice pdf got a make over:

- Subtotal is now displayed below the amount of each row and a total of all rows combined
- Discount and Tax are displayed below subtotal
- Starting balance is shown right above the total

If there's no discount, tax or starting balance then subtotal isn't shown. The new position of the subtotal, discount, tax and balance also make it much more clear on what tax is calculated.

Old invoice: 

Screen Shot 2019-06-07 at 20 48 22

New invoice: 

Screen Shot 2019-06-14 at 21 30 42

We could also opt to remove the total and rename "amount due" to total but that could get a bit confusing with the naming of the methods.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,14 +23,19 @@
         "illuminate/routing": "~5.8.0|~5.9.0",
         "illuminate/support": "~5.8.0|~5.9.0",
         "illuminate/view": "~5.8.0|~5.9.0",
+        "moneyphp/money": "^3.2",
         "nesbot/carbon": "^1.26.3|^2.0",
         "stripe/stripe-php": "^6.0",
-        "symfony/http-kernel": "^4.2"
+        "symfony/http-kernel": "^4.2",
+        "symfony/intl": "^4.3"
     },
     "require-dev": {
         "mockery/mockery": "^1.0",
         "orchestra/testbench": "^3.8",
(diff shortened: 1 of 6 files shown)


driesvints:
TIL: https://symfony.com/doc/current/components/intl.html#currencies

PR #696, opened 2019-07-08
[10.0] Payment Methods

driesvints:
These changes replace the old Sources and Token based APIs from Stripe with the new and recommended Payment Methods API. This new API will handle all card based interaction and in the future also new payment methods like IDEAL, Bancontact, etc.

Basically all card-based methods and its related class were replayed by payment method methods (wording djeez) and a new PaymentMethod class. New tests for these methods were added.

There's one legacy backwards compatibility that's maintained and that's for the defaultPaymentMethod method. This will return an instance of a Stripe\Card or Stripe\BankAccount source if no default payment method was found. This way, older apps migrating to the new Cashier release, can still use these objects to interact with their apps.

The subscription creation process now also requires a payment method instead of a Stripe token.

Developers will need to update their Stripe JS integration to make use of the new payment method apis.

Closes https://github.com/laravel/cashier/issues/649


pierrocknroll:
Hello @driesvints, my app is an API and I have to return the card informations, I used defaultCard().
Since defaultPaymentMethod can return a Stripe\Card or a PaymentMethod, do I have to do check like this or is there an other solution ?

```
$paymentMethod = $user->defaultPaymentMethod();
if (get_class($paymentMethod) === PaymentMethod::class && $paymentMethod->type === 'card' && isset($paymentMethod->card)) {
    $card = paymentMethod->card;
} else {
    $card = paymentMethod;
}
return [
    'brand'     => $card->brand,
    'country'   => $card->country,
    'exp_month' => $card->exp_month,
    'exp_year'  => $card->exp_year,
    'last4'     => $card->last4
];
```

Thanks a lot

driesvints:
@pierrocknroll unfortunately Stripe hasn't migrated the default source to the default payment methods yet so you'd indeed have to check for instanceof when retrieving it.

PR #707, opened 2019-07-19
[10.0] Improve stripe statuses

driesvints:
These changes sync the Stripe status within Cashier. This allows us to make better decisions when payments fail. There's also a new `hasIncompletePayment` method on both the billable trait and subscription model so it's easy to determine if a payment needs to be completed.

Subscriptions with an `incomplete_expired` status will be deleted from the app because they serve no purpose anymore.


pierrocknroll:
Hello @driesvints 
Since 'stripe_status' is a new column, it's not a problem for people who already have data in their table ? Do you recommend to do a script to fill the status of all current subscriptions ?

driesvints:
@pierrocknroll technically using `null` in the beginning would be fine I believe but I suggest to sync with Stripe's statuses to be 100% sure everything's running okay. Your webhooks will begin updating your subscriptions normally but it's best if you do the first sync yourself.

PR #717, opened 2019-07-25
[10.0] Re-enable proration

### LINKED ISSUE #647: Add an Prorate() function to reverse the noProrate() function
- Cashier Version: 9.3.0
- Laravel Version: 5.8+
- PHP Version: 7+
- Database Driver & Version:

### Description:
The problem with the noProrate option is that when you execute a swap right afterward a swap with noProrate() function you can't turn on Prorate.

If I have a user downgrade the plan, I do not want to prorate or give a refund, but if the user downgrades and then upgrades higher to what the originally paid during the subscription period, then I want to swap to the original plan with noProrate and then swap to a higher plan with a Prorate.

What I did to fix the issue on my end is add a new function to the Subscription model called Prorate() so that I can turn the $prorate flag to true.  If you would simply add the function to the base code it would help a lot of developers out.

denvers:
Added a prorate() function to Subscription in order to re-enable the proration after it was disabled by calling noProrate().

Fixes #647

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -312,6 +312,18 @@ class Subscription extends Model
     }
 
     /**
+     * Indicate that the plan change should be prorated.
+     *
+     * @return $this
+     */
+    public function prorate()
+    {
+        $this->prorate = true;
+
+        return $this;
+    }
+
+    /**
      * Change the billing cycle anchor on a plan change.
      *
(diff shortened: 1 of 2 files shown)


taylorotwell:
Has conflicts.

taylorotwell:
Closing pending inactivity.

PR #721, opened 2019-07-30
[9.0] Don't try and find a user when stripeId is null

nick-f:
If a payment is manually created from the Stripe dashboard, it is
possible that it will not be linked to a customer.

When this null stripeId is used to find a user record, it will find the
first customer record that does not have a stripeId in the database.

Adding an early exit will prevent this from happening, and stop these
null stripeId users being erroneously referenced in webhook handling.


driesvints:
Thanks!

PR #736, opened 2019-08-08
[10.0] Add german translations (de.json)

lanort:


taylorotwell:
I don't think we ship translation with the core. They can be shipped as community packages.

PR #739, opened 2019-08-10
[10.0] Add indexes to those columns used to lookup data in the database

ipalaus:
Digging into the code I see that `users.stripe_id` and `subscriptions.stripe_status` are both used in the query builder to query data in the database. Additionally, `subscriptions.user_id` is used to load the relationship between users and subscriptions.

I can't see why you wouldn't want these rows to be indexed for faster lookups, especially with big datasets.


taylorotwell:
Can we make a single index for user_id,status on the subscriptions table or do we query on status separately by itself?

ipalaus:
`stripe_status` is used in multiple scopes like incomplete, past due, etc. I'd probably leave them separately but in either case, it would be better than having no indexes.

PR #748, opened 2019-08-15
[10.0] syncStripeStatus trying to update incorrect status column

miticdjd:
syncStripeStatus method from Subscription model trying to update `status` column which is not correct. Correct column name is `stripe_status`.


driesvints:
Good catch. Thanks!

PR #749, opened 2019-08-15
[10.0] Update Stripe API version

josiasmontag:
Since there are [no changes that affect Cashier](https://stripe.com/docs/upgrades?since=2019-05-16#api-changelog) we could use the latest Stripe API version (2019-08-14).

### DIFF
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -15,7 +15,7 @@ class Cashier
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


taylorotwell:
@driesvints ?

driesvints:
Yeah I think it's safe to update the API now and then. But since it would require us to do minor releases and since there's no real benefit in doing so I'd advice at doing this only when there's new features we can make use of.

Issue #615, opened 2019-03-14
Paper size customizable for invoices


Mittler1:
Currently, invoices are always generated in paper format "US Letter, Portrait". Invoices should also be able to be generated in other paper formats. For example, DIN A4.

driesvints:
I can see how that would be beneficial. I've actually experienced myself wanting to use A4 instead. We could think of an easy way of passing a parameter to the view as an option to adjust the format.

driesvints:
PR merged.

PR #752, opened 2019-08-19
Paper size can be set in config file

rocramer:
The paper size can be set in the config file. 

The most parts of the world are using A4 as letter size and not US letter. (see https://en.wikipedia.org/wiki/Paper_size#/media/File:Prevalent_default_paper_size.svg) The previous default is not suitable for the most countries.

The PR is fully backwards compatible because the paper size will be set to US letter as a default if no config value is provided.


rocramer:
#615

driesvints:
Was renamed to `CASHIER_PAPER`: https://github.com/laravel/cashier/commit/cb837d13f570353b85b27bd381e8669d1fee3491

I'll add this to the docs tomorrow.

PR #763, opened 2019-08-27
[10.0] Add ability to ignore cashier routes

amranidev:
## Why

It would be better to ignore cashier routes if we want and use our own instead.

### DIFF
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -39,6 +39,13 @@ class Cashier
     public static $runsMigrations = true;
 
     /**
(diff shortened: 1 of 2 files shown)


driesvints:
But you can overwrite the routes in your `web.php` file. So not sure why you'd want to ignore the default ones?

amranidev:
@driesvints We don't want cashier service provider to register any routes, we use custom endpoints and behaviours (we don't use `web.php`), and it will keep our api docs nice and clean for us.
Why we'd keep routes that we don't want to be registered or used in anyway?

Issue #743, opened 2019-08-13
Stripe throws incomplete payment after first subscription with SCA


marcbelletre:
I'm working on updating my project with Cashier `v10.0.0-beta.2` and the new Payment Intents API for a few days now and I'm still having issues with my customers having to provide their card details and 3D secure twice. I can't say if I did something wrong or if this is a Cashier related issue so I'm asking here.  

I updated my integration following [these instruction](https://github.com/laravel/docs/blob/update-cashier-docs/billing.md#payment-methods).  

My new customers have to fill a form with their personal information and select one or many subscription plans. At the end of the form there is a Stripe card element. Before the user submits the form, I use Stripe's `handleCardSetup` method to verify the payment and then submit the form to my controller which creates the subscription(s).  

The form contains a SetupIntent key which is generated by the controller before returning the view.
```php
$setupIntent = (new User)->createSetupIntent();

return view('front.auth.register', ['setupIntent' => $setupIntent]);
```

This is my jQuery code for handling Stripe API.
```js
$('#register-form').on('submit', function(e) {
    e.preventDefault();

    // Disable submit button
    $(this).find('button[type=submit]').prop('disabled', true);

    var clientSecret = $('#setupintent-secret').val();
    var cardholderName = $('#cardholder-name').val();

    stripe.handleCardSetup(
        clientSecret, cardElement, {
            payment_method_data: {
                billing_details: {
                    name: cardholderName
                }
            }
        }
    ).then(function(result) {
        if (result.error) {
            // Inform the user if there was an error
            $('#card-errors').html(result.error.message).slideDown(200);

            // Re-enable form submission
            $('#register-form').find('input[type=submit]').prop('disabled', false);
        } else {
            // Insert the payment method ID into the form so it gets submitted to the server
            var form = $('#register-form');

            form.append($('<input type="hidden" name="stripePaymentMethod" />').val(result.setupIntent.payment_method));

            // Submit form
            form.get(0).submit();
        }
    });
});
```

I get the Stripe's 3D secure popin everytime I submit the form, but when I confirm the payment using the Complete authentication button, the SubscriptionBuilder's `create()` method throws an `IncompletePayment` exception and I have to provide all the card details again. This seems wrong as I already verified the payment with 3D secure.   

Here is the code which is handling the new subscription in my controller:
```php
$subscription = $user->newSubscription($planType, $planId);

try {
    // Create subscription
    $subscription->create($request->input('stripePaymentMethod'), [
        'email' => $user->email
[...]

PR #764, opened 2019-08-28
Set off_session parameter to true when creating a new subscription

marcbelletre:
Set `off_session` parameter to `true` when creating a new subscription to tell Stripe that the recurring payments can be made off session. Without this parameter (which is `false` by default) all other payments will throw `IncompletePayment` exceptions.

This fixes issue #743.

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -263,6 +263,7 @@ class SubscriptionBuilder
             'quantity' => $this->quantity,
             'tax_percent' => $this->getTaxPercentageForPayload(),
             'trial_end' => $this->getTrialEndForPayload(),
+            'off_session' => true,
         ]);
     }
 


skollro:
Can confirm this fixes #743, hopefully it is merged soon!

taylorotwell:
@driesvints seems to make sense to me but you can review.

hassanazimi:
You might want to add option `'confirm' => true` as well as `'off_session' => true` according to this:
https://youtu.be/95qSebQrm5E?t=2393

driesvints:
But when you're performing a charge you're always on session?

SlyDave:
Incorrect, a automatically renewing subscription (for example) will process a charge off session. As will collecting payment details on session and processing the charge when the customer isn't there (like how much of America does card payments offline still).

Adding confirm = true is interesting, as it clearly wanted required to make it all work, but seems like it should be to record the fact the user did authorise us to make the charge. One assumes without it stripe notes we charged off session without confirmation, as to what effect that has...

driesvints:
@SlyDave we're talking about the `charge` method for one-off charges, not subscriptions.

hassanazimi:
@driesvints not always. with `off_session` we can charge the customer with `$user->defaultPaymentMethod()->id`. If the card needs re-authentication then it will throw an exception which from that we need to pass `payment_intent->id` as a parameter to 

```php
$paymentIntent = PaymentIntent::retrieve($paymentIntentId)
``` 

(hold off to `payment_method->id`) and use `$paymentIntent` response in JS again with `stripe.handleCardPayment` like this (such a pain)

```
Stripe.handleCardPayment(paymentIntent.client_secret, {
				    payment_method: paymentMethodId,
			    })
			    .then(response => {
				    if (response.error) {
					    console.log(response.error.message);
				    } else {
					   // Success!
                                            // Persist do whatever you want
				    }
			    });
```

SlyDave:
My apologises, must have got confused with this being a pull request for subscriptions and not a new issue to raised for a different concern, my bad, carry on. 😁

driesvints:
@hassanazimi you can pass this with the `$options` array I guess?

hassanazimi:
@driesvints absolutely!

PR #771, opened 2019-09-04
Ensure SVG icons are visible even with a long success or error message

u01jmg3:
Before:

before

After (with a [`max-width`](https://tailwindcss.com/docs/max-width/) applied to the `span` that holds the success or error message):

after

---

ℹ️ Even with this change, the SVG icon will still differ in size depending on the length of the success or error message. An alternative to this approach would be to add a [`min-width`](https://tailwindcss.com/docs/min-width/) to the SVG icon container but Tailwind doesn't offer appropriate values out of the box.


Sti3bas:
`flex-shrink-0` class added to `svg` tag (without adding `max-width` to message's `span`) would make an icon to be the same size no matter how long the message is.

u01jmg3:
Shall re-submit with @Sti3bas suggestion

PR #773, opened 2019-09-10
remove nullable() from 'stripe_status'

Khuthaily:
removed nullable() from this guide to make it match the outcome of running the command 

`php artisan vendor:publish --tag="cashier-migrations"`


driesvints:
This won't work for people who are upgrading and already have data in their tables.

PR #784, opened 2019-09-15
[10.x] Update Stripe SDK to v7

sebdesign:
The changes from v6 to v7 seem trivial to me, so I think this can be merged into Casher v10.

## Update Stripe SDK to v7
- Replace `Charge::refund()` to `Refund::create()`
- Replace `Customer::invoices()` to `Invoice::all()`
- Replace renamed exception classes

## Fix removing default payment method
After upgrading to the v7 of the Stripe SDK, the `$customer->invoice_settings->default_payment_method` is null.

In that case, we don't need to update the Stripe customer, so we will just update the database.

The previous functionality has been kept as is.

## Lazy-load the invoice line items
According to #562, the invoice response contains only 10 line items, so we were manually fetching 1000 items.

But Stripe's pagination limit is 100 items, so that doesn't ensure that all the line items will be fetched.

Instead, we can use the `Stripe\Collection::autoPagingIterator()` which automatically fetches all the paginated results with a generator.

I was tempted to use a lazy collection here, but since this method will be called by the `Invoice::invoiceItems()` and `Invoice::subscriptions()` methods, we don't want to hit the API twice just to filter the results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /vendor
 composer.lock
 phpunit.xml
+.phpunit.result.cache
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "illuminate/view": "~5.8.0|^6.0|^7.0",
         "moneyphp/money": "^3.2",
         "nesbot/carbon": "^2.0",
-        "stripe/stripe-php": "^6.40",
+        "stripe/stripe-php": "^7.0",
         "symfony/http-kernel": "^4.3",
         "symfony/intl": "^4.3"
     },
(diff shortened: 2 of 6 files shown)


driesvints:
@taylorotwell I'll take a thorough look at this when I'm back next week.

driesvints:
@sebdesign First of all: thanks for sending this in!

I changed some things for the payment methods though. In the changes you made, the default card details could also be unset in the case when the customer was still using a default from the old legacy sources api. That would cause their default payment method to be unset. I've refactored this a little bit.

What turns out that got changed is that Stripe will now also remove the default payment method when you detach a card. That wasn't the case before. The implementation is now more simpler this way.

PR #790, opened 2019-09-25
Configure Stripe logger

sebdesign:
### Allow users to configure the Stripe library logger using Laravel's logging channels.

By default the Stripe library uses a subset of the PSR-3 logger interface to writes log messages using `error_log`.

This feature gives users the ability to use Laravel's logging channels instead, by decorating them in a `Laravel\Cashier\Logger` class, which implements Stripe's logging interface. For more advanced cases, the decorator can accept any PSR-3 implementation.

We can define the desired logging channel name in 3 ways:

**1. Using the `CASHIER_LOGGER` variable in the .env file.**

```
CASHIER_LOGGER=default
```

**2. Otherwise the `cashier.logger` configuration value can be overriden in the published `cashier.php`:**

```php
'logger' => 'default',
```

**3. Or by calling `Cashier::useLogger()` in the `register` method of a service provider:**

```php
Cashier::useLogger('default');
```

### DIFF
--- a/src/Logger.php
+++ b/src/Logger.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Laravel\Cashier;
+
+use Psr\Log\LoggerInterface;
+use Stripe\Util\LoggerInterface as StripeLogger;
+
+class Logger implements StripeLogger
+{
+    /**
+     * @var \Psr\Log\LoggerInterface
+     */
+    protected $logger;
+
+    public function __construct(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function error($message, array $context = [])
+    {
+        $this->logger->error($message, $context);
+    }
+};
--- a/tests/Unit/LoggerTest.php
+++ b/tests/Unit/LoggerTest.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace Laravel\Cashier\Tests\Unit;
+
+use Laravel\Cashier\Cashier;
+use Laravel\Cashier\Logger;
+use Laravel\Cashier\Tests\TestCase;
+use Mockery as m;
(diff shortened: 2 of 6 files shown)


tests/Unit/LoggerTest.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+namespace Laravel\Cashier\Tests\Unit;
```
driesvints: This test will need to be moved to the Integration suite because the framework is bootstrapped.

src/Logger.php
```diff
@@ -0,0 +1,27 @@
...
+use Stripe\Util\LoggerInterface as StripeLogger;
+
+class Logger implements StripeLogger
+{
+    /**
+     * @var \Psr\Log\LoggerInterface
+     */
+    protected $logger;
+
+    public function __construct(LoggerInterface $logger)
```
driesvints: Add a DocBlock to the constructor like other classes.

sebdesign:
Thank you for your feedback! I'll do these changes tonight.

Le jeu. 26 sept. 2019 à 11:06, Dries Vints <notifications@github.com> a
écrit :

driesvints:
@sebdesign took the liberty to update some stuff already. We'll get this merged in soon

sebdesign:
@driesvints thanks for pitching in! I just saw the changes you made, all looks good!

sebdesign:
@driesvints Actually I had another thought, since we can have logging channel now, what that make sense to report our exceptions (in `Cashier\Exception`) to the specified channel?

E.g. the `SubscriptionUpdateFailure` would have a method like this?

```php
public function report()
{
    Log::channel(config('cashier.logger'))->error($this->getMessage(), [
        'exception' => $this,
    ]);
}
```

driesvints:
Think it's still best to let those exceptions be handled by the user

PR #791, opened 2019-09-26
[10.x] Refactor handling of invalid webhook signatures

sebdesign:
We can rid of the `\Illuminate\Contracts\Foundation\Application` dependency in the `VerifyWebhookSignature` constructor, by manually throwing a 403 exception.

This simplifies both the implementation and the tests, and it allows us to pass the message and the original exception to the exception handler.

### DIFF
--- a/src/Http/Middleware/VerifyWebhookSignature.php
+++ b/src/Http/Middleware/VerifyWebhookSignature.php
@@ -4,19 +4,12 @@ namespace Laravel\Cashier\Http\Middleware;
 
 use Closure;
 use Illuminate\Contracts\Config\Repository as Config;
-use Illuminate\Contracts\Foundation\Application;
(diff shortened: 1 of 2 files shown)


driesvints:
On thing on my mind is people extending the class and using the app property but that seems very unlikely tbh.

sebdesign:
Yeah I doubt people would want to extend this middleware. :)

sebdesign:
@driesvints should we remove the config repository dependency as well? We are using `config('cashier...)` everywhere except from this class.

driesvints:
@sebdesign I think that was still needed for the tests but if you can simplify it then go for it.

sebdesign:
@driesvints I'll do it tonight! The goal is to simplify tests as well.

PR #793, opened 2019-09-30
[10.x] Remove config repository dependency from webhook middleware

sebdesign:
Similarly to #791, we can remove the `\Illuminate\Contracts\Config\Repository` in the `VerifyWebhookSignature`'s constructor. For consistency, the configuration is accessed with the `config()` helper like the rest of the code.

Since we don't need to mock the configuration repository anymore, the test has been moved to the integration suite, because we need to bootstrap the framework.

As a bonus, I have added tests for checking [replay attacks](https://stripe.com/docs/webhooks/signatures#replay-attacks) against the tolerance configuration setting.


driesvints:
@sebdesign I think I now realize why this was a unit test. Because we're now using the container to resolve config dependencies we can't unit test the middleware anymore. The reply attack tests are nice though. Can you maybe send in a new PR with the old implementation? Thanks!

sebdesign:
But do we really need to **unit** test the middleware? Isn't an integration test as much as valuable?

I'll send an other PR with the timing tests though, but I don't believe it's worth injecting the config repository just to be able to isolate this middleware, while the config is resolved with the helper in all the other places.

What do you think?

driesvints:
Okay, I agree after consideration.

PR #796, opened 2019-10-11
[10.x] Mark asStripe... methods internal

netpok:
Based on [this comment](https://github.com/laravel/cashier/issues/794#issuecomment-540620215) objects returned by the asStripe.. methods should not be considered part of the public api so it should be marked as internal to notify developers.

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -756,6 +756,7 @@ trait Billable
     /**
      * Get the Stripe customer for the model.
      *
+     * @internal
      * @return \Stripe\Customer
      */
     public function asStripeCustomer()
--- a/src/Invoice.php
+++ b/src/Invoice.php
@@ -320,6 +320,7 @@ class Invoice
     /**
      * Get the Stripe invoice instance.
      *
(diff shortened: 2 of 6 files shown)


driesvints:
Tbh, I don't exactly agree with this (see my comment linked above). If you're relying on functionality by the SDK then you need to add it to your composer.json file to make sure that if upstream changes are made to the used library that Composer will prevent you from upgrading to a version that your app doesn't supports.

netpok:
In my opinion exposing the underlying object and exceptions makes it part of the public API.

The [Semver FAQ entry](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api) is about packages used, but not part of the public API.

 For example cashier is using MoneyPHP to format the amount, but it is not exposed.

driesvints:
@netpok that's a valid point. Let's see what Taylor thinks.

taylorotwell:
I don't consider them internal really. Sometimes you may need the underlying object.

PR #798, opened 2019-10-12
[10.x] Allow custom mail building and redirection url in payment notification

netpok:
This pull request extends the payment notification with two features:
- It allows defining a callback to build the mailable just like the VerifyEmail and ResetPassword notification
- It allows defining an URL or a callback to provide a redirection URL after the confirmation page
- It also splits out the payment URL generation for easier overriding in case if needed.

This pull request does not change the default behavior and is totally backward compatible so Im sending it against 10.x

### DIFF
--- a/src/Notifications/ConfirmPayment.php
+++ b/src/Notifications/ConfirmPayment.php
@@ -13,6 +13,13 @@ class ConfirmPayment extends Notification implements ShouldQueue
     use Queueable;
 
     /**
+     * The callback that should be used to build the mail message.
+     *
+     * @var \Closure|null
+     */
+    public static $toMailCallback;
+
+    /**
      * The PaymentIntent identifier.
      *
      * @var string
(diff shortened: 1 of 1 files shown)


netpok:
I decided to add the notifiable to the mail build callback as it allows customization based on it and the other notifications are supporting it to.
The current parameter order is: `notifiable, amount, payment url` this makes sense to me but I'm open to suggestions.

driesvints:
Heya, thanks for sending this in. I'm a little unsure if all of this is needed though. Can you give some practical examples on how you'd use all this?

> - It allows defining an URL or a callback to provide a redirection URL after the confirmation page
> - It also splits out the payment URL generation for easier overriding in case if needed.

You can do this by publishing the notification view and overwriting the redirect.

taylorotwell:

netpok:
My intention was to allow sending the users more specified mails for example redirect them to a certain order in their history after they complete the payment. I will give it a second thought as I'm not hundred percent sure that I chose the best solution for that.

I think the custom mailable setting part stays valid as the core framework notifications allow it, how about I split that out into a separate pull request.

driesvints:
@netpok what would you use it for?

netpok:
To send customized email (we have an email template) to the customer.

While this can be done by creating a new notification for it, the same can be said of the [EmailVerification](https://github.com/laravel/framework/blob/6.x/src/Illuminate/Auth/Notifications/VerifyEmail.php) notification.

driesvints:
@netpok the reason why we added the env variable for it was indeed so you could overwrite the notification. So feel to do it like that. I don't think we need to add the exact same implementation as for the email verification class.

PR #800, opened 2019-10-18
[10.x] Retrieve payment method

geoffs3310:
I've added a method to retrieve a payment method from Stripe by its ID in order to verify that a token is valid. Since the token is generated on the front end it can be manipulated and so I needed a way to validate it in a form request before I attempt to attach it to the customer.

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -430,6 +430,17 @@ trait Billable
 
         return new PaymentMethod($this, $stripePaymentMethod);
     }
+    
+    /**
+     * Retrieve a payment method from Stripe
+     *
+     * @param  \Stripe\PaymentMethod|string  $paymentMethod
+     * @return \Laravel\Cashier\PaymentMethod
+     */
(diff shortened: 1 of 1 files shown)


driesvints:
I don't see the need for this tbh. Just use the PaymentMethod::retrieve from Stripe.

geoffs3310:
That's true of other methods in the Billable trait such as findInvoice but that made it in?

geoffs3310:
Plus you're already doing the call anyway in resolveStripePaymentMethod but because its protected it's no use to anyone. So I suggest either making that method public or merging this PR in.

driesvints:
@geoffs3310 hmm yeah I see your point. Let's try making the existing method public instead.

driesvints:
@geoffs3310 I'll send in a new PR for this in a bit.

driesvints:
Here: https://github.com/laravel/cashier/pull/801

Issue #803, opened 2019-10-20
Support for WebhookHandled event


mitulgolakiya:
# WebhookHandled Event Support

Cashier provides a great in-built way to handle webhook from Stripe. When some modification happens on a remote stripe website with the subscription it handles it and makes changes to subscriptions in the database.

## Problem/Use Case

The additional use case that we came to last week was, When any customer is canceling a Subscription, we want to send him an email to know what was the reason he canceled his subscription and get the feedback. It needs to be done from both website as well as if a user is canceling subscription form the Stripe website directly.

There can be lots of other use cases, where if something happens on Stripe, we need to notify the user after his subscription changes have been recorded to databases via Webhook. E.g.

- Notify user that he is now on a free plan and will not able to access few features
- Due to lower plan migration, he can download his data from the given link in next 7 days (in case of storage-specific plans)
etc.

## Proposed Solution

As a solution, Cashier can fire an event after successful webhook handling. A user can optionally listen for that event in code and then can read the payload and take whatever action he wanted to take by reading `$payload['type']`.

## How to listen to WebhookHandled Event

As per Laravel documentation [here](https://laravel.com/docs/6.x/events#defining-listeners), create a listener to listen to the event.

```
<?php

namespace App\Listeners;

use Laravel\Cashier\Events\WebhookHandled;

class WebhookHandledListener
{
    /**
     * Create the event listener.
     *
     * @return void
     */
    public function __construct()
    {
        //
    }

    /**
     * Handle the event.
     *
     * @param  Laravel\Cashier\Events\WebhookHandled  $event
     * @return void
     */
    public function handle(WebhookHandled $event)
    {
        // Access the payload using $event->payload...
    }
}
```

I think by this way, we can give a very clean way if someone is optionally want to take some extra actions on webhooks.

cwilby:
A solution you could use in the meantime is to override the `handleWebhook` method of `WebhookController` like so:

```php
<?php

namespace App\Http\Controllers;

use Laravel\Cashier\Http\Controllers\WebhookController as CashierController;
use Illuminate\Http\Request;

class MyWebhookController extends CashierController
{
  public function handleWebhook(Request $request)
  {
    parent::handleWebhook($request);
    event(new WebhookHandled($request));
  }
}
```

If you want the event to fire before the webhook is handled, move the dispatch above `parent::handleWebhook`

mitulgolakiya:
[...]

PR #804, opened 2019-10-20
[10.x] Added support for WebhookHandled event

mitulgolakiya:
Proposal #803


driesvints:
Maybe we can also add a `WebhookReceived` event which is fired right before the webhooks are handled? That way people can integrate with everything.

driesvints:
@mitulgolakiya I couldn't push to you repo so I took the liberty of opening a new PR. Your commit is still on it. Thanks for sending this in! 

https://github.com/laravel/cashier/pull/804

PR #816, opened 2019-11-08
[10.x] Support default TaxRates for Subscriptions

pierrocknroll:
Hello, 

This is a small PR that removes the tax_percent parameter in a Subscription creation and replaces it by the new default_tax_rates.
The Billable Trait has been updated, the taxRates that replace taxPercentage method returns nothing too and can be overriden in the User Model to return a array of TaxRate ids.

I don't think that it's necessary to add a TaxRate class into Cashier for now but let me know what you think.

### DIFF
--- a/src/Billable.php
+++ b/src/Billable.php
@@ -791,13 +791,13 @@ trait Billable
     }
 
     /**
-     * Get the tax percentage to apply to the subscription.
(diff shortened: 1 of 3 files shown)


driesvints:
Unfortunately the changes here are breaking. Feel free to re-send to master.

Issue #654, opened 2019-04-25
Add billable contract


driesvints:
An optional `Billable` contract could be added which can be used under the hood in class constructors so they always receive a correct instance. The current `Billable` trait could be renamed to a more explicit actionable naming.

PR #840, opened 2019-12-22
[11.x] Added Billable contract and renamed Billable trait to HasSubscriptions

rb1193:
Thought I'd have a crack at #654. This PR introduces a "Billable" contract that exactly matches the existing trait. I appreciate there's likely some more you wanted to do on this, and I'm also not sure if "HasSubscriptions" is the best name for the trait, but I thought I would submit and see what you think. Happy to make any necessary changes.

Was also wondering whether adding some more type hints would be desirable, where appropriate? I've assumed their absence is intentional though.

### DIFF
--- a/config/cashier.php
+++ b/config/cashier.php
@@ -51,9 +51,9 @@ return [
     | Cashier Model
     |--------------------------------------------------------------------------
     |
-    | This is the model in your application that implements the Billable trait
-    | provided by Cashier. It will serve as the primary model you use while
-    | interacting with Cashier related methods, subscriptions, and so on.
+    | This is the model in your application that implements the Billable
+    | contract provided by Cashier. It will serve as the primary model you use
+    | while interacting with Cashier related methods, subscriptions, and so on.
     |
     */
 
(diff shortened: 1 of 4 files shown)


driesvints:
@rb1193 this is great. What I was thinking which could also be cool is to split the interfaces into smaller ones like 'Chargable`, 'Subscriptionable', 'Invoiceable' (or simmilar) and then pull them together into the `Billable` contract. I can take over this pr later, thanks for putting in the initial work 👍 

Laravel indeed doesn't implements any scalar type-hints in any of its first-party libraries at this time.

rb1193:
Thanks @driesvints , you're very welcome :+1:

taylorotwell:
Dries says he will come back to this when he starts working on Cashier 11, so I am closing for now.

driesvints:
I've decided to no longer persue this path. I've been experimenting with a contract but it was quite a bit of work for little gain. Instead I believe https://github.com/laravel/cashier/issues/545 is a better path forward. We can use that model class to type hint all the places where it's used.

PR #863, opened 2020-02-10
get subscription by stripe id

edalzell:
PR'ing to master as per [Dries](https://discordapp.com/channels/297040613688475649/610866753395752968/676434076256960541).

I am building an addon on top of Cashier and part of it is sending emails on webhook events. In that listener I don't have the `Subscription` I only have the event payload.

Need a nice way to get the Subscription by Stripe id.

### DIFF
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -2,11 +2,11 @@
 
 namespace Laravel\Cashier;
 
-use Money\Money;
-use Money\Currency;
-use NumberFormatter;
 use Money\Currencies\ISOCurrencies;
+use Money\Currency;
 use Money\Formatter\IntlMoneyFormatter;
+use Money\Money;
+use NumberFormatter;
 
 class Cashier
 {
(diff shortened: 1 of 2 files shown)


taylorotwell:
I guess I'm confused... why not just do Laravel\Cashier\Subscription::where(....)... I don't know why you need to add this method here?

driesvints:
@edalzell I think we misunderstood each other. I thought you were looking to implement and easy way to re-use the `getUserByStripeId` method.

Something like we discussed with the `findCustomer` method:

```php
public static function findCustomer($stripeId)
{
    if ($stripeId === null) {
        return;
    }

    $model = config('cashier.model');

    return (new $model)->where('stripe_id', $stripeId)->first();
}
```

Which you'd then could re-use in the webhook method and elsewhere:

```php
protected function getUserByStripeId($stripeId)
{
    return Cashier::findCustomer($stripeId);
}
```

edalzell:
Ya Dries, I might re-submit w/ that as it would make my code nicer. Thanks for your patience as I learn how to PR into Cashier properly.

PR #867, opened 2020-02-17
Move getUserByStripeId to Cashier

edalzell:
There was no convenient way to grab the relevant user from a webhook event (like `customer.subscription.created` without copying the code from `WebhookController`.

Move that method to the `Cashier` class and update the calls in `WebhookController`.

Now in my webhook controller I can do:
```php
$user = Cashier::getUserByStripeId($payload['data']['object']['id']);
```

### DIFF
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -135,7 +135,7 @@ class Cashier
     }
 
     /**
-     * Get the user based on their Stripe id
+     * Get the user based on their Stripe id.
      *
      * @param string $stripeId
      * @return Billable
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -133,4 +133,21 @@ class Cashier
 
         return new static;
     }
+
+    /**
+     * Get the user based on their Stripe id
+     *
+     * @param string $stripeId
+     * @return Billable
+     */
+    public static function getUserByStripeId($stripeId)
+    {
+        if ($stripeId === null) {
+            return;
+        }
+
+        $model = config('cashier.model');
+
+        return (new $model())->where('stripe_id', $stripeId)->first();
+    }
 }
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -14,6 +14,7 @@ use Laravel\Cashier\Payment;
 use Laravel\Cashier\Subscription;
 use Stripe\PaymentIntent as StripePaymentIntent;
 use Symfony\Component\HttpFoundation\Response;
+use Laravel\Cashier\Cashier;
 
 class WebhookController extends Controller
 {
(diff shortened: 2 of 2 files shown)


src/Http/Controllers/WebhookController.php
```diff
@@ -199,23 +200,6 @@ protected function handleInvoicePaymentActionRequired(array $payload)
...
-    {
-        if ($stripeId === null) {
-            return;
-        }
-
-        $model = config('cashier.model');
-
-        return (new $model)->where('stripe_id', $stripeId)->first();
-    }
-
```
driesvints: So, like I said on Discord: it's probably best to keep this method and do the Cashier call inside of here. That way there's no BC break and people can easily overwrite the method if they need to.

src/Cashier.php
```diff
@@ -133,4 +133,21 @@ public static function keepPastDueSubscriptionsActive()
...
         return new static;
     }
+
+    /**
+     * Get the user based on their Stripe id.
+     *
+     * @param string $stripeId
+     * @return Billable
+     */
+    public static function getUserByStripeId($stripeId)
```
driesvints: Let's rename this to `findCustomer` like we discussed.

src/Cashier.php
```diff
@@ -133,4 +133,21 @@ public static function keepPastDueSubscriptionsActive()
 
         return new static;
     }
+
+    /**
+     * Get the user based on their Stripe id.
+     *
+     * @param string $stripeId
+     * @return Billable
```
driesvints:
```suggestion
     * @return \Laravel\Cashier\Billable
```

src/Cashier.php
```diff
@@ -133,4 +133,21 @@ public static function keepPastDueSubscriptionsActive()
 
         return new static;
     }
+
+    /**
+     * Get the user based on their Stripe id.
+     *
+     * @param string $stripeId
```
driesvints:
```suggestion
     * @param  string  $stripeId
```

src/Cashier.php
```diff
@@ -133,4 +133,21 @@ public static function keepPastDueSubscriptionsActive()
 
         return new static;
     }
+
+    /**
+     * Get the user based on their Stripe id.
```
driesvints:
```suggestion
     * Get the billable entity instance by Stripe ID.
```

src/Cashier.php
```diff
@@ -133,4 +133,21 @@ public static function keepPastDueSubscriptionsActive()
...
+     */
+    public static function getUserByStripeId($stripeId)
+    {
+        if ($stripeId === null) {
+            return;
+        }
+
+        $model = config('cashier.model');
+
+        return (new $model())->where('stripe_id', $stripeId)->first();
```
driesvints:
```suggestion
        return (new $model)->where('stripe_id', $stripeId)->first();
```

driesvints:
@edalzell If you take my note about putting the method inside the current we can actually do this in the current release. Can you process my feedback and send in a new PR to the 10.0 branch? Thanks!

PR #868, opened 2020-02-17
[11.x] Add subscription options + test

### LINKED ISSUE #787: Support arbitrary options to Subscription on create
The `SubscriptionBuilder::create()` method supports an $options array, which is passed on and added to the Customer. There are a couple extra options I'd like to be passing to the Stripe Subscription object as well, so I extended `SubscriptionBuilder` and added a third parameter to `create()`, `$subscriptionOptions = []`.

I merge this arbitrary array with the one created by `buildPayload()`, which then allows me to create a Subscription with options such as `backdate_start_date` without needing additional fluid functions or specific support.

I can prepare a PR for this, if this sounds like something Cashier would like to support.

keithbrink:
This is a pull request to support additional subscriptions options, and resolves #787 

I was thinking about passing $subscriptionOptions and $customer into the buildPayload function, but ultimately decided to keep the PR simple. Let me know if you'd prefer the array merge to happen in the buildPayload function.

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -204,12 +204,18 @@ class SubscriptionBuilder
      * @throws \Laravel\Cashier\Exceptions\PaymentActionRequired
      * @throws \Laravel\Cashier\Exceptions\PaymentFailure
      */
-    public function create($paymentMethod = null, array $options = [])
+    public function create($paymentMethod = null, array $options = [], $subscriptionOptions = [])
     {
         $customer = $this->getStripeCustomer($paymentMethod, $options);
 
+        $payload = array_merge(
+            ['customer' => $customer->id],
+            $this->buildPayload(),
+            $subscriptionOptions,
+        );
+
         $stripeSubscription = StripeSubscription::create(
-            ['customer' => $customer->id] + $this->buildPayload(),
+            $payload,
             $this->owner->stripeOptions()
         );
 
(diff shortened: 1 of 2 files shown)


src/SubscriptionBuilder.php
```diff
@@ -204,12 +204,18 @@ public function add(array $options = [])
      * @throws \Laravel\Cashier\Exceptions\PaymentActionRequired
      * @throws \Laravel\Cashier\Exceptions\PaymentFailure
      */
-    public function create($paymentMethod = null, array $options = [])
+    public function create($paymentMethod = null, array $options = [], $subscriptionOptions = [])
```
driesvints: Let's also rename `$options` here to `$customerOptions`.

tests/Integration/SubscriptionsTest.php
```diff
@@ -615,4 +615,17 @@ public function test_subscriptions_with_tax_rates_can_be_created()
 
         $this->assertEquals([self::$taxRateId], [$stripeSubscription->default_tax_rates[0]->id]);
     }
+
+    public function test_subscriptions_with_options_can_be_created()
+    {
+        $user = $this->createCustomer('subscriptions_with_options_can_be_created');
+
+        $backdate_timestamp = now()->subMonth()->getTimestamp();
```
driesvints:
```suggestion
        $backdateStartDate = now()->subMonth()->getTimestamp();
```

driesvints:
@keithbrink thanks!

PR #872, opened 2020-02-20
Add @throws declaration to methods on Billable which can throw Payment exceptions

ValCanBuild:
In the documentation https://laravel.com/docs/6.x/billing#payments-requiring-additional-confirmation it's stated the the methods `charge` `invoice` and `invoiceFor` on a user can throw `IncompletePayment` exceptions but the methods documention wasn't marked as such.

This PR just adds those `@throws` tags to the methods.


driesvints:
Thanks!

PR #873, opened 2020-02-20
Ability to pass in additional options when creating a subscription

MarGul:
I had an issue. When migrating users to use cashier I needed to pass in [proration_behaviour](https://stripe.com/docs/api/subscriptions/create#create_subscription-proration_behavior) and set it to `none` when creating a subscription. 

With this PR I can now do this:

```php
$user->newSubscription('main', 'plan')->create($paymentMethod, [], ['proration_behaviour' => 'none']);
```

I've seen that you should have tests if a PR should be approved but when looking through the tests I did not find that you are testing `$options` in the create method today so I skipped tests here as well. Please let me know if, and also how, I should incorporate tests.

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -198,17 +198,18 @@ class SubscriptionBuilder
      *
      * @param  \Stripe\PaymentMethod|string|null  $paymentMethod
      * @param  array  $options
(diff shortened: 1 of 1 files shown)


driesvints:
This has already been merged on master and will be available in the upcoming v11 release.

PR #886, opened 2020-02-28
[10.x] Re-enable proration

driesvints:
Re-send of https://github.com/laravel/cashier/pull/717

This will allow to re-enable prorating when doing multiple swaps in one request. Also adds tests for proration.

Closes https://github.com/laravel/cashier/issues/647

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -397,6 +397,18 @@ class Subscription extends Model
     }
 
     /**
+     * Indicate that the plan change should be prorated.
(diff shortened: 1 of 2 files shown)


driesvints:
@denvers I wasn't able to recover your commit unfortunately. But thanks for your work on this 👍 

Btw I think you made a wrong assertion on the end of your original test where you expected an invoice of zero. I've updated the test to expect the correct `-1000` value.

PR #908, opened 2020-04-02
syncStripe method in Subscription class

chrisfritsche:
New method "syncStripe" in Subscription class, to perform full synchronization. Motivation: active scope from subscriptions check status but also "ends_at" value (and ignore canceled status).

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -156,6 +156,23 @@ class Subscription extends Model
     }
 
     /**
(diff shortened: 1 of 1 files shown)


driesvints:
No plans on adding this.

driesvints:
You should configure your Stripe webhooks to keep your susbcriptions up to date: https://laravel.com/docs/7.x/billing#handling-stripe-webhooks

PR #916, opened 2020-04-16
[11.x] Re-add tax percentage

driesvints:
After discussing with @themsaid we decided to re-add the old `taxPercentage` way of defining tax rates to make it easier for people to migrate. Tax Rates will remain the primary way of defining taxes and we won't re-add the `taxPercentage` to the documentation. All `taxPercentage` calls are also marked as deprecated.


jbrooksuk:
Test for Dries.

PR #926, opened 2020-04-27
[11.x] Fix add and remove plan

driesvints:
Fixes https://github.com/laravel/cashier/issues/922

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -720,7 +720,7 @@ class Subscription extends Model
 
         $this->unsetRelation('items');
 
-        if ($this->items()->count() > 1) {
+        if ($this->hasSinglePlan()) {
             $this->fill([
                 'stripe_plan' => null,
                 'quantity' => null,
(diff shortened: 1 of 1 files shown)


Nuranto:
@driesvints I don't know the code, I just saw your commit, so maybe I will say something stupid.
But should'nt we create the single plan item if not exist before setting plan to null ?

Something like this : 

```
if ($this->hasSinglePlan()) {
           if ($this->items()->count() ==0 ) { // Can occurs after upgrading from version 10 to 11
                   $this->items()->create([
                           'stripe_id' => $subscription->stripe_id,
                           'stripe_plan' => $subscription->stripe_plan,
                           'quantity' => $subscription->quantity,
                   ]);
            }
            $this->fill([
                'stripe_plan' => null,
                'quantity' => null,
            ])->save();
}
```

driesvints:
@Nuranto no that's not needed. Your webhooks will take care of that.

PR #935, opened 2020-05-12
Use sendInvoice() when collection method is 'send_invoice'

eckelarsson:
In some edge cases (like mine) we have clients that can't use a credit card, but we still want to have all the data in Stripe. So we create a subscription with collection_method = 'send_invoice' and the manually sent the invoices to be paid, once we received payment we just change the status manually in Stripe.

```
$model->invoice(['collection_method' => 'send_invoice']);
$subscription->invoice(['collection_method' => 'send_invoice']);
```

When you use this collection method you can't apply `$stripeInvoice->pay()` function, you need to use the `$stripeInvoice->sendInvoice();`

When using the `$stripeInvoice->pay()` on when a colletion_method is 'send_invoice' it throws PaymentFailure, 'send_invoice' it doesn't require a payment method.

The only two options in 'collection_methods' are charge_automatically (default) or 'send_invoice' therefore using a simple if-else statement.

https://stripe.com/docs/api/invoices/create#create_invoice-collection_method

### DIFF
--- a/src/Concerns/ManagesInvoices.php
+++ b/src/Concerns/ManagesInvoices.php
@@ -76,7 +76,11 @@ trait ManagesInvoices
             /** @var \Stripe\Invoice $invoice */
             $stripeInvoice = StripeInvoice::create($parameters, $this->stripeOptions());
 
-            $stripeInvoice = $stripeInvoice->pay();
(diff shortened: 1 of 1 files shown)


driesvints:
@eckelarsson thanks for the pr. We've merged your commit through another pr. This will be in next Tuesday's release.

PR #937, opened 2020-05-13
Expand option to asStripeSubscriptionItem method

eckelarsson:
Enable the option to expand the SubscriptionItem::retrieve as you can with the asStripeSubscription method.

My use case is to be able to expand the plan's product to retrieve the name of the related product. I think this will be helpful when listing multiple subscription items on the "billing page" and want to display the actual product names related to each subscription item. The response from Stripe API is only the product id, which is not that descriptive for the user.

### DIFF
--- a/src/SubscriptionItem.php
+++ b/src/SubscriptionItem.php
@@ -183,12 +183,13 @@ class SubscriptionItem extends Model
     /**
      * Get the subscription as a Stripe subscription item object.
      *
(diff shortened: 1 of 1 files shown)


driesvints:
Unfortunately we cannot make these changes on a minor release because the method signature was changed. I've pushed your change up on the master branch and it'll be available in the next major release. Thanks

PR #941, opened 2020-05-18
[12.x] Rename all subscription boolean method with `is` as prefix

### LINKED ISSUE #940: Subscription scopes doesn't work
- Cashier Version: 11.2
- Laravel Version: 7.0
- PHP Version: 7.3.11
- Database Driver & Version: sqlite 3.29 & mysql 8

### Description:
Hi, I found a bug using Subscription's scopes. 
All scopes are overwritten by scope-named method (active overwrite scopeActive.. etc.)
Other users experienced similar issues: https://laracasts.com/discuss/channels/general-discussion/error-when-calling-scope-deprecated-non-static-method?page=0

### Steps To Reproduce:
```php
\Laravel\Cashier\Subscription::active()->count()
```
results:
```
ErrorException
Non-static method Laravel\Cashier\Subscription::active() should not be called statically
```

I tried also to call `\Laravel\Cashier\Subscription::scopeActive()->count()` but the result is simliar.
It is the same for ALL scope methods.

One suggestion (that brokes back-compatibility, but these methods never worked) is to rename all scope-named method with a `is` prefix. (eg. `isActive` `isIncomplete`).

I found a tricky fix here https://laravel.io/forum/non-static-method-should-not-be-called-statically-for-scopes

joke2k:
Resolve #940 breaking back-compatibility
The method names now are clear and you can use scope methods like Laravel standard.

This pull request is against the master because it requires a Major version (#which-branch)[https://laravel.com/docs/contributions#which-branch]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased](https://github.com/laravel/cashier/compare/v11.2.4...master)
 
+### Changed
+- Renamed all Subscription boolean method with a `is` prefix. ([#940](https://github.com/laravel/cashier/issues/940))
 
 ## [v11.2.4 (2020-05-08)](https://github.com/laravel/cashier/compare/v11.2.4...v11.2.4)
 
--- a/src/Concerns/ManagesSubscriptions.php
+++ b/src/Concerns/ManagesSubscriptions.php
@@ -34,7 +34,7 @@ trait ManagesSubscriptions
 
         $subscription = $this->subscription($name);
 
-        if (! $subscription || ! $subscription->onTrial()) {
+        if (! $subscription || ! $subscription->isOnTrial()) {
             return false;
         }
 
--- a/src/Concerns/ManagesSubscriptions.php
+++ b/src/Concerns/ManagesSubscriptions.php
@@ -26,9 +26,9 @@ trait ManagesSubscriptions
      * @param  string|null  $plan
      * @return bool
      */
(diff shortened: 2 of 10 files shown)


driesvints:
I'm not entirely opposed to this. Although there's quite some breaking changes with this.

@taylorotwell btw tests are failing because forks don't have access to our Stripe test key so you can ignore that.

joke2k:
I'm sorry about that... I'm a hellish stubborn. 😅

Actually, testing over stripe test API floods my staging app.

taylorotwell:
Don't understand this PR at all.

joke2k:
What is too difficult to understand for you?

The `Subscription::active` (and similar methods) matches with `Subscription::scopeActive` usage from the framework.
So if I use `Subscription::active()->get()` it raises error. 
I hope that's clear

driesvints:
@joke2k I've had another talk with the team about this and we've decided that this isn't worth the breaking change.

You can still easily use the scopes like so: `Subscription::query()->active()->get()`.

I'll send in a PR to the docs to document them like this.

joke2k:
@driesvints it makes sense, my PR would introduce a big change in the package.

I hope that https://github.com/joke2k/docs/commit/004e82d860fc830a4704f64bef995fee429fb099 could help you to update doc.

driesvints:
@joke2k thanks, that's helpful indeed: https://github.com/laravel/docs/pull/6069

PR #970, opened 2020-07-06
[feature] Customizable Subscription model

rennokki:
I am trying to build some features on top of `Laravel\Cashier\Subscription` model, and I find a good idea to make it customizable via the config file.

### DIFF
--- a/config/cashier.php
+++ b/config/cashier.php
@@ -61,6 +61,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Subscription Model
+    |--------------------------------------------------------------------------
+    |
+    | This is the model in your application that is used to track the
+    | subscriptions. It is used to run certain actions regarding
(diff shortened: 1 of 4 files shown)


driesvints:
You can already customize it by overwriting the relationship on the billable model:

```
class User extends Model
{
    use Billable;

    /**
     * Get all of the subscriptions for the Stripe model.
     *
     * @return \Illuminate\Database\Eloquent\Relations\HasMany
     */
    public function subscriptions()
    {
        return $this->hasMany(YourCustomModel::class, $this->getForeignKey())->orderBy('created_at', 'desc');
    }
}
```

rennokki:
What about the `SubscriptionItem` ?

https://github.com/laravel/cashier-stripe/pull/970/files#diff-8f775a106146d885bf9b5f47ba6928eaR41

driesvints:
Overwrite the items method in your custom model.

PR #991, opened 2020-09-10
Move user model to directory

d8vjork:
As Laravel 8 moved the models to its own directory.

Only downside is that old installations will need to keep the exported config or specify the model on the env.


taylorotwell:
This is a breaking change on 12.x. We would have to make this change on a new major release - but for now you can just configure the model in your .env file.

Issue #637, opened 2019-04-11
Stripe Checkout 


driesvints:
The new Stripe Checkout product provides an easy for to get started with payments without having to do a manual setup. The trade-off is that the payment process is completely styled and moved over to Stripe. 

~We'll only implement this in v10 if it gets a stable release soon. Otherwise can pick this up later.~ Checkout is out of beta so it'll probably make it into the new release.

https://stripe.com/docs/payments/checkout

aldovincenti:
Thanks. Do you plan to implement this feature before September 14? As from that date, Cashier would not be available for a [business based in the EEA](https://stripe.com/docs/strong-customer-authentication), without this feature.

driesvints:
@aldovincenti we're aware and it's the whole reason we're now working on v10. See the milestone: https://github.com/laravel/cashier/milestone/1

If Checkout gets a stable release soon it'll probably make it into v10 and otherwise before September.

driesvints:
Checkout is out of beta so it'll probably make it into the new release.

driesvints:
Progress: https://github.com/laravel/cashier/pull/652

driesvints:
Since Stripe still hasn't updated their API with the necessary changes in order to let this move forward we're taking this out of the next release. We'll have to wait and see when Stripe makes the necessary changes listed in the PR.

NeoMarine:
Is there a way for me to begin using Stripe Checkout with Laravel Cashier currently? It would save me a lot of time worrying about building an on-site form, handling the 3D secure stuff, etc.

You can quite easily redirect to the Stripe Checkout - it just appears there's just no webhook set for "checkout.session.completed" - is there a lot more that needs to be added?

It seems that perhaps if I added my own webhook for "checkout.session.completed" I would be able to then call something like _**$user->newSubscription('main', 'premium')->create($token);**_ - currently when I use the Stripe Checkout, it returns to the success_url as I define it, but no subscription is added.

I'd appreciate any help here. If I need to simply not use checkout then I suppose I have no choice, but it would be great...

driesvints:
@NeoMarine No. Please see the concerns on the WIP PR for Checkout in Cashier on why Checkout is currently incompatible with Cashier: https://github.com/laravel/cashier/pull/652

driesvints:
PR was closed for now. We can later pick up the work from the closed PR.

cihantas:
Any chance this will ship in v12?

driesvints:
[...]

PR #652, opened 2019-04-23
[10.0] Implement Stripe Checkout

driesvints:
This PR adds functionality for Stripe checkout. At the moment it'll initiate a new server side session so we can add the checkout to an existing user. Unfortunately I just discovered that this isn't implemented by Stripe yet: https://stripe.com/docs/payments/checkout/server#span-classstepoptionalspan-using-existing-customers

> You may specify an existing Customer object for Checkout to use when making one-time payments (this does not yet work with plans and subscriptions). 

We'll have to wait until Stripe implements this before continuing with subscriptions. We can already see at implementing checkout for single charges though.

## Todo

- [ ] Subscribing to plans
- [ ] Swapping plans
- [ ] Single charges
- [ ] Webhooks (Payment Fulfillments)
- [ ] Style button
- [ ] Tests
- [ ] Write documentation
- [ ] Prefill session id in success link: https://stripe.com/docs/payments/checkout/fulfillment#webhooks

## Concerns

Below I'll try to list all concerns I encounter during development.

### Existing customers can't yet be passed along when subscribing to a plan

This is vital for Cashier to work so until this gets implemented on Stripe's end this PR is basically blocked.

Screen Shot 2019-04-26 at 20 48 58

https://stripe.com/docs/payments/checkout/migration#client-subscriptions

### Trial days seem to work differently for Checkout sessions ✅

Current api for subscriptions: https://stripe.com/docs/api/subscriptions/create#create_subscription-trial_end

Screen Shot 2019-04-26 at 20 55 06

Api when creating Checkout sessions: https://stripe.com/docs/api/checkout/sessions/create#create_checkout_session-subscription_data-trial_end

Screen Shot 2019-04-26 at 20 54 41

This basically means that you're always required to give either a 48h period or 1 day period when subscribing through Checkout. This seems unwanted and weird. I'll check this over with Stripe.

Update 30/04: Stripe confirmed that the 48h limitation is in place for the following reason:

> Checkout Sessions are active for 24 hours after their creation and within that timeframe, your customer can complete the payment at any time. We require the trial start least 48 hours in the future so that there is still at least a 1 day trial if your customer pays at the end of the 24 hours.

I've implemented this here: https://github.com/laravel/cashier/commit/ff9946d4c88cbefdb483829da3a0f18612e57854

### Unclear how to upgrade or downgrade using Checkout

[...]

PR #1007, opened 2020-09-29
[12.x] Stripe Checkout Support

driesvints:
This PR adds functionality for Stripe checkout and is a continuation of https://github.com/laravel/cashier-stripe/pull/652. All concerns on the original PR have been resolved since.

Docs: https://github.com/laravel/docs/pull/6465

### Usage

Checking out a single priced product:

```php
$checkout = Auth::user()->checkout('price_xxx');

return view('checkout', compact('checkout'));
```

Checking out a single priced product with a specific quantity:

```php
$checkout = Auth::user()->checkout(['price_xxx' => 5]);

return view('checkout', compact('checkout'));
```

Checking out multiple priced products and optionally assign quantities to some:

```php
$checkout = Auth::user()->checkout(['price_xxx' => 5, 'price_yyy']);

return view('checkout', compact('checkout'));
```

Checking out a single priced product and allow promotional codes to be applied:

```php
$checkout = Auth::user()->allowPromotionCodes()->checkout('price_xxx');

return view('checkout', compact('checkout'));
```

Checking out a new $12 priced product (this will create a new product in the dashboard):

```php
$checkout = Auth::user()->checkoutCharge(1200, 'T-shirt');

return view('checkout', compact('checkout'));
```

Checking out a new $12 priced product together with a Price ID product:

```php
$checkout = Auth::user()->checkout([
    'price_xxx', 
    [
        'price_data' => [
            'currency' => Auth::user()->preferredCurrency(),
            'product_data' => [
                'name' => 'T-Shirt',
            ],
            'unit_amount' => 1200,
        ],
    ],
]);

return view('checkout', compact('checkout'));
```

Check out a subscription:

```php
$checkout = Auth::user()->newSubscription('default', 'price_xxx')->checkout();

return view('checkout', compact('checkout'));
```

Check out a subscription and allow promotional codes to be applied:

```php
$checkout = Auth::user()->newSubscription('default', 'price_xxx')
    ->allowPromotionCodes()
    ->checkout();

return view('checkout', compact('checkout'));
```

Then place any of these in a view:

```php
{!! $checkout->button() !!}
```

#### Button Styling

By default a generated checkout button will have the following styling:

Screenshot 2020-10-05 at 18 00 30

It's however easy to style this. Pass either a custom `style` or `class` attribute:

```php
{!! $checkout->button('Buy', ['class' => 'text-white bg-blue-500 p-4']) !!}
```

### Todos

- [x] Subscribing to plans
- [x] Charging products
- [x] Single charges
- [x] Webhooks
- [x] Style button
- [x] Tests
- [x] Write documentation

Closes https://github.com/laravel/cashier-stripe/issues/637

### DIFF
--- a/resources/views/checkout.blade.php
+++ b/resources/views/checkout.blade.php
@@ -15,7 +15,7 @@
 
         checkoutButton.addEventListener('click', function () {
             // When the customer clicks on the button, redirect them to Checkout.
-            Stripe('{{ config('cashier.key') }}').redirectToCheckout({
+            Stripe('{{ $stripeKey }}').redirectToCheckout({
                 sessionId: '{{ $sessionId }}'
             }).then(function (result) {
                 // If `redirectToCheckout` fails due to a browser or network
--- a/resources/views/checkout.blade.php
+++ b/resources/views/checkout.blade.php
@@ -1,0 +1,27 @@
+<button
+    id="checkout-{{ $sessionId }}"
+    role="link"
+    style="background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em"
+>
+    {{ $label }}
+</button>
+
+<div id="error-message"></div>
+
+<script>
+    const checkoutButton = document.getElementById('checkout-{{ $sessionId }}');
+
+    checkoutButton.addEventListener('click', function () {
+        // When the customer clicks on the button, redirect them to Checkout.
+        Stripe('{{ config('cashier.key') }}').redirectToCheckout({
+            sessionId: '{{ $sessionId }}'
+        }).then(function (result) {
+            // If `redirectToCheckout` fails due to a browser or network
+            // error, display the localized error message to your customer
+            // using `result.error.message`.
+            if (result.error) {
+                document.getElementById('error-message').innerText = result.error.message;
+            }
+        });
+    });
+</script>
--- a/resources/views/checkout.blade.php
+++ b/resources/views/checkout.blade.php
@@ -9,19 +9,21 @@
 <div id="error-message"></div>
 
 <script>
-    const checkoutButton = document.getElementById('checkout-{{ $sessionId }}');
+    (() => {
+        const checkoutButton = document.getElementById('checkout-{{ $sessionId }}');
 
-    checkoutButton.addEventListener('click', function () {
-        // When the customer clicks on the button, redirect them to Checkout.
-        Stripe('{{ config('cashier.key') }}').redirectToCheckout({
-            sessionId: '{{ $sessionId }}'
-        }).then(function (result) {
-            // If `redirectToCheckout` fails due to a browser or network
-            // error, display the localized error message to your customer
-            // using `result.error.message`.
-            if (result.error) {
-                document.getElementById('error-message').innerText = result.error.message;
-            }
+        checkoutButton.addEventListener('click', function () {
+            // When the customer clicks on the button, redirect them to Checkout.
+            Stripe('{{ config('cashier.key') }}').redirectToCheckout({
+                sessionId: '{{ $sessionId }}'
+            }).then(function (result) {
+                // If `redirectToCheckout` fails due to a browser or network
+                // error, display the localized error message to your customer
+                // using `result.error.message`.
+                if (result.error) {
+                    document.getElementById('error-message').innerText = result.error.message;
+                }
+            });
         });
-    });
+    })()
 </script>
--- a/resources/views/checkout.blade.php
+++ b/resources/views/checkout.blade.php
@@ -1,7 +1,8 @@
 <button
     id="checkout-{{ $sessionId }}"
     role="link"
-    style="background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em"
+    style="{{ isset($style) && ! isset($class) ? $style : 'background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em' }}"
+    @isset($class) class="{{ $class }}" @endisset
 >
     {{ $label }}
 </button>
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -75,6 +75,7 @@ class Checkout
         return View::make('cashier::checkout', array_merge([
             'label' => $label,
             'sessionId' => $this->session->id,
+            'stripeKey' => config('cashier.key'),
         ], $options));
     }
 
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -1,0 +1,88 @@
+<?php
+
+namespace Laravel\Cashier;
+
+use Illuminate\Support\Facades\View;
+use Stripe\Checkout\Session;
+
+class Checkout
+{
+    /**
+     * @var \Illuminate\Database\Eloquent\Model
+     */
+    protected $owner;
+
+    /**
+     * @var \Stripe\Checkout\Session
+     */
+    protected $session;
+
+    /**
+     * Create a new Checkout instance.
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $owner
+     * @param  \Stripe\Checkout\Session  $session
+     * @return void
+     */
+    public function __construct($owner, Session $session)
+    {
+        $this->owner = $owner;
+        $this->session = $session;
+    }
+
+    /**
+     * Get the Checkout Session ID.
+     *
+     * @return string
+     */
+    public function sessionId()
+    {
+        return $this->session->id;
+    }
+
+    /**
+     * Begin a new Checkout Session.
+     *
+     * @param  \Illuminate\Database\Eloquent\Model  $owner
+     * @param  array  $sessionOptions
+     * @param  array  $customerOptions
+     * @return \Laravel\Cashier\Checkout
+     */
+    public static function create($owner, array $sessionOptions = [], array $customerOptions = [])
+    {
+        $customer = $owner->createOrGetStripeCustomer($customerOptions);
+
+        $session = Session::create(array_merge([
+            'customer' => $customer->id,
+            'mode' => 'payment',
+            'success_url' => route('home').'?checkout=success',
+            'cancel_url' => route('home').'?checkout=cancelled',
+            'payment_method_types' => ['card'],
+        ], $sessionOptions), Cashier::stripeOptions());
+
+        return new static($customer, $session);
+    }
+
+    /**
+     * Get the View instance for the button.
+     *
+     * @param  string  $label
+     * @param  array  $options
+     * @return \Illuminate\Contracts\View\View
+     */
+    public function button($label = 'Check out', array $options = [])
+    {
+        return View::make('cashier::checkout', array_merge([
+            'label' => $label,
+            'sessionId' => $this->session->id,
+        ], $options));
+    }
+
+    /**
+     * @return \Stripe\Checkout\Session
+     */
+    public function asStripeCheckoutSession()
+    {
+        return $this->session;
+    }
+}
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -8,17 +8,21 @@ use Stripe\Checkout\Session;
 class Checkout
 {
     /**
+     * The Stripe model instance.
+     *
      * @var \Illuminate\Database\Eloquent\Model
      */
     protected $owner;
 
     /**
+     * The Stripe Checkout Session instance.
+     *
      * @var \Stripe\Checkout\Session
      */
     protected $session;
 
     /**
-     * Create a new Checkout instance.
+     * Create a new Checkout Session instance.
      *
      * @param  \Illuminate\Database\Eloquent\Model  $owner
      * @param  \Stripe\Checkout\Session  $session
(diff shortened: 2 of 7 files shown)


resources/views/checkout.blade.php
```diff
@@ -0,0 +1,27 @@
...
+    role="link"
+    style="background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em"
+>
+    {{ $label }}
+</button>
+
+<div id="error-message"></div>
+
+<script>
+    const checkoutButton = document.getElementById('checkout-{{ $sessionId }}');
```
cjavilla-stripe, line 12: one thing we've run into is a few users will try to put multiple buttons on one page. I wonder if it makes sense to wrap this bit in an iife to keep the scope of `checkoutButton` inside of the single view?

```html
<script>
(() => {
  const checkoutButton = doc...
  //
  
)()
</script>
```

cjavilla-stripe, line 12: One thing we've run into is folks putting multiple buttons on the same page. I wonder if for this bit if you want to wrap this in an iife? That way the scope of checkoutButton is isolated to each instance of the checkout view on the page? 

```html
<script>
  (() => {
    const checkoutButton = doc...
   ///
  })()
```

src/Checkout.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Get the View instance for the button.
+     *
+     * @param  string  $label
+     * @param  array  $options
+     * @return \Illuminate\Contracts\View\View
+     */
+    public function button($label = 'Check out', array $options = [])
```
chasenyc, line 67: nitpick but it seems like everywhere it is referred to as `Checkout` not `Check out`

src/SubscriptionBuilder.php
```diff
@@ -99,7 +99,7 @@ public function __construct($owner, $name, $plans = null)
     public function plan($plan, $quantity = 1)
     {
         $options = [
-            'plan' => $plan,
+            'price' => $plan,
```
chasenyc, line 102: this does not affect how subscriptions are currently created? this seems confusing to me, is this taking in a plan id or the price of the plan?

driesvints: It's the same thing. Both accept a price or a plan id. Stripe has provided BC for this to ease the transition to their new Prices API. But Checkout requires it to be "price". All tests still succeed btw which means this works.

src/Checkout.php
driesvints: https://twitter.com/stauffermatt/status/1036953677704163330

driesvints:
Wrote the docs for this today: https://github.com/laravel/docs/pull/6465

src/Concerns/PerformsCharges.php
```diff
@@ -57,4 +58,49 @@ public function refund($paymentIntent, array $options = [])
...
+     * @param  string  $price
+     * @param  int  $quantity
+     * @param  array  $sessionOptions
+     * @param  array  $customerOptions
+     * @return \Laravel\Cashier\Checkout
+     */
+    public function checkoutProduct($price, $quantity = 1, array $sessionOptions = [], array $customerOptions = [])
+    {
+        return Checkout::create($this, array_merge([
+            'line_items' => [[
```
juancho48, line 100: why the double array [[ ?

cjavilla-stripe: line items is an "array of hashes" or in this case an array of associative arrays. So the outer `[` is for the "array of" and the second `[` is for the first "hash" in the list.

juancho48: got it thanks!

driesvints:
We've updated `checkout($amount, $productName, ...)` and `checkoutProduct($productPrice, ...)` to `checkout($productPrice, ...)` & `checkoutCharge($amount, $productName, ...)` since you'd mostly want to checkout predefined products instead of creating ad-hoc products which pollute your dashboard.

JhumanJ:
Hi, 
I've been following this issue for some time, and I'm looking for it to be merged. I needed this functionality so I built something similar on my own while this gets merged. However, I'm having an issue, so I was curious to know if you were also facing it. 

Whenever I create a subscription via checkout, Stripe send to webhooks, one for the subscription creation (with a status `incomplete`) and one for the subscription update (setting the status to `active`). These two hooks usually arrive at the exact same time (or very close). Therefore when the update hook handle method runs, it occurs that the subscription was not created yet in my database. Because of that, the subscription will not be active in my DB causing some obvious logical isssues...

Any thoughts on that ?

cjavilla-stripe:
@JhumanJ The order that webhook events are delivered isn't guaranteed. There are a couple options. (1) listen for the `checkout.session.completed` event. That'll let you know when the session completes and you'll know the Subscription is started. (2) push the events into a queue, then process later in the background in order.

JhumanJ:
@cjavilla-stripe I understand the issue, but I'm worried that everyone will be facing it with the current state of this pull request...

src/Http/Controllers/WebhookController.php
```diff
@@ -55,6 +55,49 @@ public function handleWebhook(Request $request)
...
+                    $subscription->items()->create([
+                        'stripe_id' => $item['id'],
+                        'stripe_plan' => $item['plan']['id'],
+                        'quantity' => $item['quantity'],
+                    ]);
+                }
+            }
+        }
+
+        return new Response('Webhook Handled', 200);
```
jasonvarga, line 98: You could use `$this->successMethod();`

driesvints:
Note to self: https://stripe.com/docs/api/checkout/sessions/create#create_checkout_session-discounts

dannywp (commented):
I don't see any code that would update the User's default payment method. Am I missing something or has it been overlooked?

driesvints:
@dannywp the current `customer.updated` webhook already handles that in Cashier.

src/Http/Controllers/WebhookController.php
driesvints: Good catch thanks.

dannywp:
@driesvints correct, but only if the stripe customer has a default payment method. I have been implementing stripe hosted checkout for subscriptions and it seems like the card that is attached to the customer is not set as the default payment method.

driesvints:
@dannywp last time I tried it it worked for me. You need to explicitly check it as the new default payment method.

driesvints:
@JhumanJ heya, you're gonna have to help me walk through on how to recreate your situation. I haven't had a situation yet where the original subscription was set to incomplete first and the subscription update event was triggered. Can you share the specifics on how to recreate your use case?

cjavilla-stripe:
@dannywp to jump in with a bit more context. When creating a subscription with Checkout, a new PaymentMethod is created and a new Subscription is created. You can either create the Checkout session with or without a customer object. If no customer object is provided, then a new customer will also be created.

When collecting recurring payments for a subscription, Stripe will determine which payment method to use for the invoice by looking in a few places. 

First, we'll look at the Subscription itself to see if it has a [default_payment_method](https://stripe.com/docs/api/subscriptions/object#subscription_object-default_payment_method) set. This is where the payment method is set in the case of Checkout in `subscription` mode. This payment method will be used when collecting future payments for this subscription unless the default payment method on the subscription is changed or removed.

Second, we'll look at the Subscriptions `default_source`. This is a deprecated field and is not commonly set, is not used by Checkout, and I'd recommend against using `default_source` anymore.

Third, if there is no `default_payment_method` or `default_source` set on the Subscription, we'll next look at the customer's `invoice_settings.default_payment_method`. This is not set by Checkout by default. This is a field that is _only_ for Billing (Subscriptions and Invoices) and does not act as a default for one time payments with PaymentIntents. If you're creating a *second* subscription for a customer who's first subscription was created with Checkout, you might want to make an API call to set the customer's `invoice_settings.default_payment_method` to be equal to the first subscription's default_payment_method.

Fourth, we'll look at the customer's `default_source` also deprecated, and also not recommended.

If you are collecting payments outside of the automatic recurring payments that result from the Subscription created with Checkout, I would recommend storing the ID of the default_payment_method from the original subscription as a field in your db, then using that for future payments. (I don't know whether or not this is built into the existing Checkout flow with cashier :D) 

Hope this helps 👍

stevergill:
@driesvints I'm involved with a charity trying to solve the need for foodbanks and was thinking of waiting for this PR to be merged before implementing the "donate" function on their website.

Are you hoping to have this merged in the next few weeks, or is it more likely to be next year? Completely understand either way. If I had the expertise I would offer to help.

driesvints:
@StephenGillCoop We're currently evaluating the feature internally. I don't have an ETA, sorry.

driesvints:
@JhumanJ I again tried out scenarios with test cards from https://stripe.com/docs/testing but don't encounter any unexpected bug from happening. I think it's best that you be really specific about the scenario that you encounter if I'm to recreate it.

andrebreia:
I just had the same issue as @JhumanJ. In my case my `customer.subscription.created` webhook failed (had an error in my code) and `customer.subscription.updated` succeeded, which caused the subscription status to be `incomplete` when the created webhook was re-sent.

driesvints:
@andrebreia we can't account for errors in your code. If you don't have an error in your code and the webhooks work properly then I think the feature works as expected.

andrebreia:
@driesvints That's fair. Just thought that there might be situations where the webhook could fail.

BKirev:
What if we want to have multiple checkout buttons? Like this:

Checkout demo

driesvints:
@BKirev like this:

```
$checkout = Auth::user()->checkout('price_xxx');
$otherCheckout = Auth::user()->checkout('price_yyy');

return view('checkout', compact('checkout', 'otherCheckout'));
```

abbasali:
Maybe I am missing something, but how to add multiple line items (prices) for single priced products in the same checkout? From what I see you can only add one product/price to the checkout. I haven't checked the code and purely going by the docs.

driesvints:
@abbasali thanks for that input. I've pushed some changes which now allow `checkout` to be used with multiple items. I've updated the examples in the PR description to show how you can use them. Does that seem okay to you?

These changes has removed the "quantity" parameter on the `checkout` method.

abbasali:
@driesvints The changes look good, thanks. However, it doesn't support the [inline pricing object](https://stripe.com/docs/api/checkout/sessions/create#create_checkout_session-line_items-price_data). Would be nice to have inline `price_data` support so that dynamic prices like shipping costs can be added to the checkout.

driesvints:
@abbasali it does have support for that though:

```php
$checkout = Auth::user()->checkout([
     [
          'price_data' => [...]
     ],
]);

return view('checkout', compact('checkout'));
```

abbasali:
Ok cool. Wasn't in the PR docs so thought it isn't there :). Thanks.

driesvints:
@abbasali added an example. It's a bit more cumbersome to add a custom item together with existing price ID products. I'm not sure what we can do there to improve the API.

abbasali:
@driesvints What if the API is changed to something like..

```php
$checkout = $user->checkout()
    ->addItem('price_xxx', 1)
    ->addItem('price_yyy')
    ->addItem(['price_data' => [...]])
    ->create();
```
or something along those lines?

Otherwise, the current API looks good. Yes, the custom price is a bit cumbersome but if documented and explained properly - it should be fine.

matheuscandido:
Can I use this branch on my Laravel project while it hasn't been merged to 12.x?

driesvints:
@matheuscandido purely on your own risk as we might merge and delete this branch at any time. Forking the package and requiring it as a VCS might be the better choice.

driesvints:
@abbasali yeah I'm just gonna leave it as is I think.

abbasali:
> Can I use this branch on my Laravel project while it hasn't been merged to 12.x?

@matheuscandido I also needed the Checkout feature. So what I did is created a custom trait `PerformsCheckout` and copied the `PerformsCharges::checkout()` and related methods in that trait. Then created a class `Checkout` and copied `src/Checkout.php`'s code into it and used this in `PerformsCheckout` instead of cashier's `Checkout`.

Then in User model, I used the `PerformsCheckout` trait and I am now able to do `$user->checkout('price_xxx');`

Lastly, for the button HTML, I created a custom view (again copied the code from this branch).

I didn't need the subscription logic so left out all other modified files/logic from this branch and just the above 3 files made it work for me. Later when this branch goes into 12.x, I will simply remove the trait `PerformsCheckout` from my User model and everything (theoretically) should work with 12.x's new checkout code (from `PerformsCharges` cashier trait).

amirkoklan:
is there any timeline for this branch to be merged?

driesvints:
No sorry

driesvints:
This will be in next week's release!

abbasali:
Great news! Thanks for the update.

BouleyAustin:
What's the best way to integrate Rewardful using the Laravel Cashier version of Stripe Checkout?

Screen Shot 2021-02-03 at 4 53 08 PM

naumanzchaudhry:
Fixed priced items don't seem to make it to database? I am using `$user->checkout($priceId, [ // options])`? Any pointers? Have all the webhooks setup mentioned in the docs properly

driesvints:
Please open an issue

PR #1021, opened 2020-10-30
[11.x] Fail url checking when url is invalid (#984)

nickrupert7:
This PR is to copy a bug fix that was committed to the `12.x` branch into the `11.x` branch.
See https://github.com/laravel/cashier-stripe/pull/984/

The current version of `laravel/spark-aurelius` (11.0) currently requires `"laravel/cashier": "^11.0"`, in which this bug is still present.

I am aware that the Laravel Contribution guide specifies that all bug fixes should only be committed in the latest stable branch. However, since the latest stable branch of `laravel/spark-aurelius` requires an older version of `laravel/cashier`, I believe that it would be appropriate to perform this patch on the older branch.

### DIFF
--- a/src/Http/Middleware/VerifyRedirectUrl.php
+++ b/src/Http/Middleware/VerifyRedirectUrl.php
@@ -20,7 +20,9 @@ class VerifyRedirectUrl
     {
         $redirect = $request->get('redirect');
 
-        if ($redirect && parse_url($redirect)['host'] !== $request->getHost()) {
+        $url = parse_url($redirect);
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
@driesvints can you verify this fix is correct? I don't mind getting this into 11.x

driesvints:
Yep looks fine thanks!

driesvints:
Released v11.3.1 with this fix: https://github.com/laravel/cashier-stripe/releases/tag/v11.3.1

PR #1025, opened 2020-11-03
Allow lookup for paymentMethods of a specific type

christoph-kluge:
This addresses laravel#1024.

This change still keeps `type => card` as default type parameter but does also allow to use a custom type passed via `$parameters`.


driesvints:
There is a lot more to be done than this before we can offer proper support for other payment types. Like updating the `payments.blade.php` file, making sure that webhooks come in properly, updating other payment methods, etc...

PR #1032, opened 2020-11-24
[12.x] Add withPromotionCode method

driesvints:
This adds a `withPromotionCode` method to the subscription builder in addition to the `withCoupon` method in order to provide for Stripe's new promotion codes. When using this method, the API ID for the promotion code needs to be passed, not the promotion code itself.

```php
$subscription = Auth::user()->newSubscription('default', 'price_xxx')
    ->withPromotionCode('promo_xxx')
    ->create($paymentMethod);
```

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -58,11 +58,18 @@ class SubscriptionBuilder
     protected $billingCycleAnchor = null;
 
     /**
-     * The coupon code being applied to the customer.
+     * The coupon being applied to the subscription.
      *
      * @var string|null
      */
     protected $coupon;
+
+    /**
+     * The promotion code being applied to the subscription.
+     *
+     * @var string|null
+     */
+    protected $promotionCode;
 
     /**
      * The metadata to apply to the subscription.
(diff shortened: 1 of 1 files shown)


Naveedali1234:
how can we retrieve the promotion code using promo code?

driesvints:
@Naveedali1234 I don't understand your question, can you explain a bit more thoroughly?

Naveedali1234:
okay let me explain it. 

i am trying to verify promotion code if it exist in stripe dashboard or not. so when a user enter a promo code in text box for example: 7OIzSX3E

now i will need to retrieve this promo code from stripe dashboard if it exist or not, so i fetch it in this way: `$promotion_code = \Stripe\PromotionCode::retrieve($request_input('promo_code'), []);`

but this input is only the code and this input needs ID of the promotion code. 

How can i achieve it? 

i am sorry if i am asking something different, but i am just trying to fetch the response of it.

driesvints:
You can use the Stripe API like you posted above. I myself would just hardcode a list of promotion codes and their ids in the app itself. If you need something more dynamic, use the Stripe API.

jfvoliveira:
Hey there @driesvints .

Couldn't we use a promo code as well when swaping subscriptions?

```php
$user->subscription('default', $newPriceId)
   ->withPromotionCode($code)
   ->swapAndInvoice($newPriceId);
```

This doesn't work. From what I've been able to check, it only works in newSubscriptions.
Shouldn't it also work with swaps or am I thinking wrong?

Thanks

driesvints:
@jfvoliveira done: https://github.com/laravel/cashier-stripe/pull/1230

PR #1039, opened 2020-12-16
[12.x] Support for Metered Billing

### LINKED ISSUE #555: Support for Metered billing and creating UsageRecord events
Good day.

I'm working on a project that bills users with Stripe Metered Billing (https://stripe.com/docs/billing/subscriptions/metered-billing)

Are there any plans for implementing this functionality? From what I can tell it would involve allowing new subscription creation with a 0 value and a method for creating UsageRecords (possibly similar to the current `tab` method?)

Thanks,
Nic.

benjamindoe:
This PR adds support for [Metered Billing](https://stripe.com/docs/billing/subscriptions/metered-billing).

Closes #555

## Usage

Reporting usage from the subscription:

```php
$user->subscription('default')->reportUsage(); // defaults to quantity of 1
$user->subscription('default')->reportUsage(10);
$user->subscription('default')->reportUsageFor('pro-plan', 10);
```

Reporting usage from a subscription item:

```php
$item = $$user->subscription('default')->findItemOrFail('pro-plan');

$item->reportUsage();
$item->reportUsage(10);
```

Stripe gives you the ability to update a usage record at a particular timestamp. The `reportUsage` & `reportUsageFor` methods accept a Carbon object to specify the timestamp:

```php
$subscription->reportUsage(100, now()->subHour());
$subscription->reportUsageFor('pro-plan', 100, now()->subHour());

// Or report for a specific subscription item...
$item->reportUsage(100, now()->subHour());
```

## Todos

- [x] Reporting usage
- [x] Fetching usage records
- [ ] [Billing tresholds](https://stripe.com/docs/billing/subscriptions/metered-billing/thresholds
)
- [ ] Swapping behavior
- [ ] Stripe Checkout integration
- [ ] Cancelling subscriptions
- [ ] Tests

## Related Stripe Documentation

- https://stripe.com/docs/billing/subscriptions/metered-billing
- https://stripe.com/docs/billing/subscriptions/metered
- https://stripe.com/docs/api/usage_records
- https://stripe.com/docs/api/subscriptions/update#update_subscription-items-clear_usage


benjamindoe:
This PR needs quantity to be nullable. It could be that this means this is a breaking change and needs to be sent to master. Obviously, Doctrine DBAL is required to change table columns. It may be that this PR does away with the migration that changes the column to nullable and in the upgrade guide, it asks users to set quantity as nullable to be able to use metered billing.

benjamindoe:
@driesvints It would be good to get your thoughts on migrations. Should there be a migration to set the quantity to nullable or should we change the old migration and ask people to create their own migration in an upgrade guide?

driesvints:
Thanks for this PR. At the moment I don't have time to review this. I'm thinking this will be for January at the earliest.

benjamindoe:
@driesvints Do you know how I can get the CI working for this PR? Unit tests work locally fine and I've added a Stripe API key to my secrets just in case but still can't seem to find an API key to test with

driesvints:
Heya. There's some extra info here: https://github.com/laravel/cashier-stripe/blob/12.x/.github/CONTRIBUTING.md

Your test keys can be found here: https://dashboard.stripe.com/test/apikeys

benjamindoe:
Hey, thanks for getting back. Yeah I set that up and it works locally just fine. I just cant get these CI tests working

driesvints:
@benjamindoe you need to set the env variables like the secret key in the secrets settings of your fork: https://github.com/benjamindoe/cashier-stripe/settings/secrets/actions

benjamindoe:
I've opened this PR for review, however I'm still unable to get the CI tests to use the secret key that I have set in my secrets. I definitely have the STRIPE_SECRET in the secrets

image

driesvints:
@benjamindoe very strange. Anyway, I'd just leave it be as I'm very close now to giving this a review. Hopefully this or next week.

driesvints:
@benjamindoe I've started reviewing this and there's a couple of things I want to address right away/ask about:

1. At the moment I don't see a reason to add a new table for usage records. I don't see a benefit to keep these records in the database as Stripe already stores them. To retrieve current usage we can consult the upcoming invoice as described here: https://stripe.com/docs/billing/subscriptions/metered-billing#retrieving-usage. Did you have specific reasons to keep these records in a new table?
2. I see that whenever you swap plans you clear previous usage records. While the Stripe docs also often refer to `clear_usage` parameter, it strikes me as odd that previous usage is cleared when plans are swapped. The customer has consciously used part of the subscription so they should be billed for it unless the vendor explicitly wants to clear that previous usage. Can you explain a little more why you implemented it like that?

Overall this PR looks like a great start for this feature. I'm still contemplating on wether the change to the `quantity` column is a breaking one or not. We might need to send this in to master but for now let's focus on the questions above so we can clear out the main path for this PR. I'll await your answers 👍

driesvints:
Converting this to a draft until it's finished so it doesn't ends up in Taylor's list of PRs to review.

benjamindoe:
1. Only reason for this was due to stripe's recommendation that the app should track "your own usage" (https://stripe.com/docs/billing/subscriptions/metered-billing#reporting-usage). Otherwise, I agree. I don't see much benefit apart from having the usage to hand without having to send another request to the Stripe API.
2. Stripe seems to fall over when you try and swap an item without clearing usage records. You can see this if you remove the `clear_usage` and run the unit tests.

driesvints:
> Only reason for this was due to Stripe's recommendation that the app should track "your own usage" 

That's true. But I think they mean that more as the way the app tracks usage for emails for example. That falls more into the domain of the app itself and not the billing part. We've not implemented a way to keep track of Stripe's usage records, not the way the app keeps track of a customer's usage. So I'd remove all of that from the PR.

> Perhaps it's down to the app rather than this package to track usage. If this is the case, maybe we drop the migration.

Fully agree here.

> Stripe seems to fall over when you try and swap an item without clearing usage records. You can see this if you remove the clear_usage and run the unit tests.

I'll try that out.

Can you maybe update the PR to remove everything for tracking usage records? I'll give it a more thorough review after that.

benjamindoe:
Removed interval usage record tracking, added a new function to retrieve a summary of the usage records although not sure if it's entirely necessary.

driesvints:
I've started updating this PR a bit. I'll get back once I'm done.

driesvints:
I didn't entirely get to finish this, I'll continue on monday. Feel free to leave feedback on the refactors I made if you like.

driesvints:
@benjamindoe heya. Maybe it's best that you don't make any changes to this branch anymore just to avoid that I overwrite anything. I should have this finished soon I hope.

driesvints:
@benjamindoe I've updated the main description of this PR so people aren't confused about the changed code. I'll keep updating that when I add new things to the PR so please don't edit it yourself.

driesvints:
@benjamindoe sorry, I've sent in a new PR here so I can make the tests pass: https://github.com/laravel/cashier-stripe/pull/1048

PR #1053, opened 2021-02-05
[12.x] Fix return type for invoice line items

sebdesign:
`invoiceLineItemsByType` actually returns an array and not a collection.


driesvints (approved):

driesvints:
Thanks!

PR #1055, opened 2021-02-09
Add PaymentMethod type parameter to ManagesPaymentMethods

### LINKED ISSUE #507: Billable::cards does not return bank accounts
As you can see below, when pulling the sources back it's being filtered by ['object' => 'card']. This means that it will not return bank accounts. The issue with this method is that it is consumed by the deleteCards() method. This means that bank accounts are not removed by the deleteCards method, and if a customer deletes an association to a bank account, and attempts to re-add it, an exception is thrown that the bank account already exists for that customer.

``` php
    $stripeCards = $this->asStripeCustomer()->sources->all(
        // Only cards are returned here, bank accounts are not pulled
        ['object' => 'card'] + $parameters
    );
```

simonhir:
I added a `$type` parameter to have the possibility to retrieve other payment methods from Stripe than `card`.

Should fix #507

This probably should go on `master` as the `card_brand`-field now contains `card` instead of `visa` or `mastercard`.

Would be nice if somebody could say if this is more a Minor or a Major version feature and so on if I should provide the backwards compatibility.

### DIFF
--- a/src/Concerns/ManagesPaymentMethods.php
+++ b/src/Concerns/ManagesPaymentMethods.php
@@ -36,22 +36,24 @@ trait ManagesPaymentMethods
     }
 
     /**
-     * Determines if the customer currently has at least one payment method.
+     * Determines if the customer currently has at least one payment method of the given type.
      *
+     * @param String $type
      * @return bool
      */
-    public function hasPaymentMethod()
+    public function hasPaymentMethod($type = 'card')
     {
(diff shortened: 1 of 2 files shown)


driesvints:
Hey @simonhir. This indeed needs to go to master because other than the change in behavior this also breaks the method signatures. Feel free to send this to master and I'll try to take a look soon.

PR #1056, opened 2021-02-09
[13.x] Add PaymentMethod type parameter to ManagesPaymentMethods

### LINKED ISSUE #507: Billable::cards does not return bank accounts

simonhir:
I added a `$type` parameter to have the possibility to retrieve other payment methods from Stripe than card.

Should fix #507

I changed the `card_brand` field a bit to hold the PaymenMethods type. So instead of only `visa`, `mastercard` or `Bank Account` it now contains for example `card` or `sepa_debit`. The `card_last_four` field is filled if the given type has the `last4` attribute. So it would probably be a good idea to rename it to `card_type`.

### DIFF
--- a/src/Concerns/ManagesPaymentMethods.php
+++ b/src/Concerns/ManagesPaymentMethods.php
@@ -36,22 +36,24 @@ trait ManagesPaymentMethods
     }
 
     /**
-     * Determines if the customer currently has at least one payment method.
+     * Determines if the customer currently has at least one payment method of the given type.
      *
+     * @param String $type
      * @return bool
      */
-    public function hasPaymentMethod()
+    public function hasPaymentMethod($type = 'card')
     {
-        return $this->paymentMethods()->isNotEmpty();
+        return $this->paymentMethods([], $type)->isNotEmpty();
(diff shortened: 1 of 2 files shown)


driesvints:
Putting this in draft so it doesn't ends up in Taylor's PR's to review. I'll review this soon, thanks.

driesvints:
Thanks @simonhir. I finally found time to dig into this one. I opened a new PR with adjustments I made but which include your commits. This also makes the test suite passing: https://github.com/laravel/cashier-stripe/pull/1074

PR #1058, opened 2021-02-09
[12.x] Allow Stripe dashboard subscriptions

driesvints:
Allows subscriptions to be created from the Stripe dashboard. The `newSubscriptionName` allows users to configure a different default key when overriding the `WebhookController`. 

The only limitation with this PR is that it won't work in situations where you have more than one type of subscription and need to differentiate between the names.

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -76,7 +76,7 @@ class WebhookController extends Controller
                 }
 
                 $subscription = $user->subscriptions()->create([
-                    'name' => $data['metadata']['name'] ?? $this->newSubscriptionName(),
+                    'name' => $data['metadata']['name'] ?? $this->newSubscriptionName($payload),
                     'stripe_id' => $data['id'],
                     'stripe_status' => $data['status'],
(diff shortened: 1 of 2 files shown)


LukeTowers:
@driesvints is there anything preventing the newSubscriptionName() method from evaluating the payload to identify the products / prices in the subscription provided by stripe and using an internal mapping of that to dynamically return a subscription name for Cashier to use?

driesvints:
@LukeTowers heya. Not really. Once you have the payload you can do whatever you want with it I think.

LukeTowers:
@driesvints awesome, thanks! I thought that would be the case, was just asking because of your extra note in the original PR: 

> The only limitation with this PR is that it won't work in situations where you have more than one type of subscription and need to differentiate between the names.

driesvints:
@LukeTowers ah yeah sorry. In case you have multiple subscription types, for example, a tennis and a gym subscription, you can't derive the name from the subscription alone because there's no way to set metadata when subscribing a user from the dashboard. But you could indeed derive it from the prices like you said, that's a good shout.

PR #1060, opened 2021-02-17
[12.x] Ability to fill using a \Stripe\Source object

benjamindoe:

### DIFF
--- a/src/Concerns/ManagesPaymentMethods.php
+++ b/src/Concerns/ManagesPaymentMethods.php
@@ -9,6 +9,7 @@ use Stripe\Card as StripeCard;
 use Stripe\Customer as StripeCustomer;
 use Stripe\PaymentMethod as StripePaymentMethod;
 use Stripe\SetupIntent as StripeSetupIntent;
+use Stripe\Source as StripeSource;
 
 trait ManagesPaymentMethods
 {
@@ -232,7 +233,7 @@ trait ManagesPaymentMethods
     /**
      * Fills the model's properties with the source from Stripe.
      *
-     * @param  \Stripe\Card|\Stripe\BankAccount|null  $source
+     * @param  \Stripe\Card|\Stripe\BankAccount|\Stripe\Source|null  $source
      * @return $this
      *
      * @deprecated Will be removed in a future Cashier update. You should use the new payment methods API instead.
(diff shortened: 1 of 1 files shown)


driesvints:
Hey @benjamindoe, I want to decline this one as I'd rather encourage people to use the Payment Methods API instead of the deprecated and obsolete Sources API.

benjamindoe:
That's fine. The issue I was having was with legacy customers. The method to update default payment method from stripe wouldn't import a Source object, which is what is returned from `defaultPaymentMethod()`.

What do you think about getting that method to convert a legacy source to PaymentMethod? Don't know how possible that is though.

driesvints:
@benjamindoe unfortunately the only way that I know of to convert a source object to a payment method is to let people re-add their payment details with the payment methods API. It's one of the most sensitive operations with Stripe so that's probably the reason why the API's are so strict.

It's indeed true that the `defaultPaymentMethod` currently still returns the `default_source` for BC reasons but we could be removing that at some point in the future.

benjamindoe:
Coming back to this PR, I've found that Stripe dashboard still uses source API under the hood for some reason. Someone from our team added a card on behalf of a customer using Stripe dashboard and our database didn't update with the webhook due to not being able to fill from `Stripe\Source`. It seems like it's still core to Stripe so probably should be supported.

As a workaround I've extended the `Billable` trait to include my own `fillSourceDetails` method.

driesvints:
I'm sorry @benjamindoe but I don't agree here. Stripe is clearly deprecating the sources API over payment intents/payment methods. See https://stripe.com/docs/payments/older-apis

And:

Screenshot 2021-02-23 at 10 59 20

It may very well be that Stripe still uses the sources API under the hood but I believe if that's the case that they'll change that sooner rather than later. I don't think it's wise to continue supporting it. The only references left in Cashier (like the default payment method) are for BC reasons.

benjamindoe:
Thanks for the update, that message doesn't appear unless you're on a EU/GB account so I missed it. I'll concede it's not a core part of Stripe then.

It's a bit weird that the Stripe dashboard still uses an API that's been deprecated for over a year. I'd like to see what Stripe has to say on this as it's a completely valid use case for admins to be adding cards on behalf of customers via the Stripe dashboard. Cashier won't update customer payment methods while they still use Source API.

driesvints:
@benjamindoe I pinged one of my contacts at Stripe to see why that is.

driesvints:
@benjamindoe I've spoken at lengths about this with my contacts at Stripe and confirmed that Stripe itself still uses the sources API when adding new payment methods through the dashboard.

However, from what I understand they recognise this pain point and plan to address this hopefully soon. In light of that we'll not be making any changes here, sorry. But still thanks for bringing this up @benjamindoe.

I also want to note that for the next Cashier update we'll be removing all code related to the deprecated sources API and will only be using the payment methods API going forward: https://github.com/laravel/cashier-stripe/pull/1077

PR #1067, opened 2021-02-19
[12.x] Allow model swapping

driesvints:
Allows people to swap the base models in the same fashion as Passport & Sanctum.


benjamindoe:
I've hacked this into my app, didn't realise this was in the pipeline

driesvints:
@benjamindoe only decided today to introduce this.

PR #1068, opened 2021-02-19
[12.x] requiresConfirmation status helper method

benjamindoe:
In some cases, users needs to confirm a payment or setup intent. This is probably due to the Stripe customer account having a card added through the Stripe dashboard (Unconfirmed, only speculating).

You can currently just do
```php
$payment->status === \Stripe\PaymentIntent::STATUS_REQUIRES_CONFIRMATION
```

However, this `requiresConfirmation` method would bring this status inline with the other statuses

### DIFF
--- a/src/Payment.php
+++ b/src/Payment.php
@@ -76,7 +76,6 @@ class Payment
         return $this->paymentIntent->status === StripePaymentIntent::STATUS_REQUIRES_ACTION;
     }
 
-
     /**
      * Determine if the payment needs to be confirmed.
      *
--- a/src/Payment.php
+++ b/src/Payment.php
@@ -76,6 +76,17 @@ class Payment
         return $this->paymentIntent->status === StripePaymentIntent::STATUS_REQUIRES_ACTION;
     }
 
+
+    /**
+     * Determine if the payment needs to be confirmed.
+     *
+     * @return bool
+     */
+    public function requiresConfirmation()
+    {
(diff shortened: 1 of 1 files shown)


driesvints:
Can you add a method for `requires_capture` as well?

driesvints:
Would you say we need to update the `public function validate()` method as well to throw an exception? This could maybe have some implications on existing apps so unsure myself.

benjamindoe:
it depends, you could argue it's a bug fix as it was missing before and it's a state that the payment intent could potentially find itself in

benjamindoe:
On a side note, it's very very similar to "Requires action" and I'm actually using the same Stripe code to confirm

driesvints:
Right. We'll also need to update the payments view then if we're going to add these to the validate method. I'm going to leave this in draft for now so I can take a thorough look at all the implications soon. Sorry if this is gonna take a bit longer.

benjamindoe:
Sure no problem. I figured it might be quick as it's marked as "Optional" in the documentation

driesvints:
Hey @benjamindoe. Sorry this took longer. I'm already going to approve this PR for 12.x. I've been working on integrating the status checks for the `validate` method locally but I'm gonna send that in to 13.x as that could be a breaking change potentially. Thanks!

driesvints (approved):

driesvints:
@benjamindoe Here's a PR that'll also implement this check on the payment page: https://github.com/laravel/cashier-stripe/pull/1094

PR #1070, opened 2021-02-21
[12.x] Fix styles overwriting checkout button when class is set

### LINKED ISSUE #1069: Checkout button, cannot change style if class is set, and if class is set, styles overwrite class.
- Cashier Version: Latest
- Laravel Version: Latest
- PHP Version: Latest

### Description:

The `isset($style) && ! isset($class)` will not work to style buttons with classes. Because if the class isset, the if statement will return the default style, overwriting the classes. Shouldn't it be: `isset($style) && isset($class) ? $style` so the default styles don't overwrite the classes? https://raw.githubusercontent.com/laravel/cashier-stripe/a2b9716eb7b4aedb197315b61e942318441fb09e/resources/views/checkout.blade.php

I could send a PR to fix this?

```

<button
    id="checkout-{{ $sessionId }}"
    role="link"
    style="{{ isset($style) && ! isset($class) ? $style : 'background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em' }}"
    @isset($class) class="{{ $class }}" @endisset
>
    {{ $label }}
</button>

```

The below code will still return the default style, even if I add `'style' => ''`
```
{{ $checkout->button('Pay For Job', [
    'class' => 'bg-blue-100 text-sm px-4 py-2',
]) }}
```

### Steps To Reproduce:

Please see above

natecorkish:
Fixes #1069 

If the class is set, the styles would overwrite the class styles, because we're showing default styles because class isset.

Now you can do something like:

```
{{ $checkout->button('Pay For Job', [
    'class' => 'bg-blue-100 text-sm px-4 py-2',
    'styles' => 'other styles here that don't clash with classes.'
]) }}
```

### DIFF
--- a/resources/views/checkout.blade.php
+++ b/resources/views/checkout.blade.php
@@ -1,7 +1,7 @@
 <button
     id="checkout-{{ $sessionId }}"
     role="link"
-    @isset($style) style="{{ $style }}" @elseif(! isset($style) && ! isset($class)) style="background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em {{ $style }}" @endisset
+    @isset($style) style="{{ $style }}" @elseif(! isset($style) && ! isset($class)) style="background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em" @endisset
     @isset($class) class="{{ $class }}" @endisset
 >
     {{ $label }}
--- a/resources/views/checkout.blade.php
+++ b/resources/views/checkout.blade.php
@@ -1,7 +1,7 @@
 <button
     id="checkout-{{ $sessionId }}"
     role="link"
-    style="{{ isset($style) && ! isset($class) ? $style : 'background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em' }}"
+    style="{{ isset($style) && isset($class) ? $style : 'background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em' }}"
     @isset($class) class="{{ $class }}" @endisset
 >
     {{ $label }}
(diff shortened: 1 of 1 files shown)


taylorotwell:
Doesn't this just create the reverse problem? The inline styles will potentially interfere with classes people are injecting?

taylorotwell:
To me the behavior I would expect would be to only show these default styles if NO style is passed **and** NO class is passed by the developer. If either of those things are passed we can't safely add these default styles. /cc @driesvints

taylorotwell:
In other words:

```
style="{{ ! isset($style) && ! isset($class) ? 'background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em' : ''; }}"
```

natecorkish:
> In other words:
> 
> ```
> style="{{ ! isset($style) && ! isset($class) ? 'background-color:#6772E5;color:#FFF;padding:8px 12px;border:0;border-radius:4px;font-size:1em' : ''; }}"
> ```

That makes sense. Updated the commit.

driesvints:
The way the current PR is shaped would cause the `$style` attribute not to be applied at all. You probably still need to add it in the same fashion as `$class`.

In any case we need to be aware that this is technically a breaking change for anyone using the current behavior. If anyone was either adding styling with `$class` or `$style` and also relying on the default styling to be present they'll see their buttons broken.

natecorkish:
> The way the current PR is shaped would cause the `$style` attribute not to be applied at all. You probably still need to add it in the same fashion as `$class`.
> 
> In any case we need to be aware that this is technically a breaking change for anyone using the current behavior. If anyone was either adding styling with the `$class` not set and the `$style` set and also relying on the default styling to be present they'll see their buttons broken.
> 
> I do agree that the new behavior makes more sense.

Updated behaviour. Can now do something like:


```
{{ $checkout->button('Pay For Job', [
    'styles' => 'color: #fff'
]) }}

```

or

```
{{ $checkout->button('Pay For Job', [
    'class' => 'text-red-50'
]) }}

```

If style isset, it'll use default style and custom style, If class isset, will use that.

driesvints:
In any case in the next major update we'll be able to implement a proper Blade component since we're dropping Laravel 6 support.

PR #1077, opened 2021-02-26
[13.x] Remove legacy sources support

driesvints:
This PR removes all support for the legacy sources API. Cashier will from now on only use the new Payment Methods API.


taylorotwell:
Looks like this one has conflicts.

driesvints:
@taylorotwell fixed.

Issue #1079, opened 2021-03-03
pass parameters to missingMethod in the method handleWebhook in WebhookController


rabol:
The method signature of looks like this:

```
 protected function missingMethod($parameters = [])
    {
        return new Response;
    }
```

it is only called once in the WebhookController, in the handleWebhook() method, but no parameters is passed to the method so if I create my own web hook controller and override the missingMethod to capture what is actually missing there is no information.

driesvints:
Thanks. I've sent in a PR for this here: https://github.com/laravel/cashier-stripe/pull/1080

Please try to fill out the issue template next time, thanks.

PR #1080, opened 2021-03-04
[12.x] Fix missing method on WebhookController

driesvints:
Fixes https://github.com/laravel/cashier-stripe/issues/1079

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -52,7 +52,7 @@ class WebhookController extends Controller
             return $response;
         }
 
-        return $this->missingMethod();
+        return $this->missingMethod($payload);
     }
 
     /**
--- a/tests/Unit/WebhookControllerTest.php
+++ b/tests/Unit/WebhookControllerTest.php
@@ -52,6 +52,7 @@ class WebhookControllerTest extends TestCase
         Event::assertNotDispatched(WebhookHandled::class);
 
         $this->assertEquals(200, $response->getStatusCode());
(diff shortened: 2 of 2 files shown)


rabol:
Seems like the code is lost

driesvints:
@rabol it's not? https://github.com/laravel/cashier-stripe/blob/12.x/src/Http/Controllers/WebhookController.php#L55

rabol:
https://github.com/laravel/cashier-stripe/pull/1080/commits/8199445995a3599ced1cfa6fa190658b3a080bc7

look at the method:

```
public function missingMethod($parameters = [])
    {
        return new Response('Missing event type: '.$parameters['type']);
    }
```

in the original PR

and then look at the current version

```
/**
     * Handle calls to missing methods on the controller.
     *
     * @param  array  $parameters
     * @return \Symfony\Component\HttpFoundation\Response
     */
    protected function missingMethod($parameters = [])
    {
        return new Response;
    }

```

driesvints:
@rabol you're pointing to code in a test. The method in the `missingMethod` method of the `WebhookController` is suppose to be empty.

rabol:
well, the original bug was that a parameter was passed to the missingMethod, but never used
and the 'fix' was that parameter should be used, e.g. in the Response og Log::info()
 
if the 'missingMethod' don 'report' that something is missing, then it is difficult to figure out which method is missing

rabol:
aha.... go it, one have to create a new controller, and override to capture it

driesvints:
Yes, the parameters is simple there to let anyone who wants to override the method and make use of it.

PR #1085, opened 2021-03-08
[12.x] Subscription Update Webhook bugfix

fschirinzi:
Reproduction:
1. Add subscription to user
2. Cancel subscription with custom date on Stripe Dashboard

Expected:
- `ends_at` is set as in the stripe subscription model property `cancel_at`

Outcome:
- `ends_at` property in subscription model is null

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -160,7 +160,7 @@ class WebhookController extends Controller
                         $subscription->ends_at = $subscription->onTrial()
                             ? $subscription->trial_ends_at
                             : Carbon::createFromTimestamp($data['current_period_end']);
-                    } elseif($data['cancel_at']) {
+                    } elseif ($data['cancel_at']) {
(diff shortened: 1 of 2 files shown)


fschirinzi:
Do I have to set the Stripe Secret in my fork to run the tests?

driesvints:
> Do I have to set the Stripe Secret in my fork to run the tests?

No, that's okay. Atm GitHub Actions doesn't share repo secrets.

driesvints:
This lgtm but I'm going to confirm this with someone from Stripe first. I can't find any docs other than the API docs about `cancel_at`.

driesvints:
Verified that this is indeed the correct solution. Thanks @fschirinzi!

driesvints:
Fyi, I had to add this `isset` because it broke some other tests: https://github.com/laravel/cashier-stripe/commit/3184afce07b65986d7aaa6f570b9ac44f502303c

PR #1091, opened 2021-03-11
[12.x] Verify trial ends in the future

benjamindoe:
Not sure if this constitutes a breaking change or not, I've sent to 12.x though.

Stripe will throw a `Stripe\Exception\InvalidRequestException` with the following message if the trial_end field is in the past.

> Invalid timestamp: must be an integer Unix timestamp in the future

This PR checks to see if the `trialExpires` property is in the future before setting trial fields. It could be argued that it should throw an exception, however I think it's much more elegant and removes a bit of boilerplate checking from the consuming app if this check is done for the user within the package.

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -321,7 +321,7 @@ class SubscriptionBuilder
             'stripe_status' => $stripeSubscription->status,
             'stripe_plan' => $isSinglePlan ? $firstItem->plan->id : null,
             'quantity' => $isSinglePlan ? $firstItem->quantity : null,
-            'trial_ends_at' => ! $this->skipTrial ? $this->trialExpires : null,
+            'trial_ends_at' => ! $this->skipTrial && $this->trialExpires->isFuture() ? $this->trialExpires : null,
             'ends_at' => null,
         ]);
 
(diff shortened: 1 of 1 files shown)


driesvints:
Hmmm. Difficult to say. On one hand I'd argue that setting a specific time through `trialUntil` is done deliberately by the developer and if they set a value in the past it's most likely to be a bug in their app. Therefor an exception being throw is probably the better choice.

However, one could equally argue that setting a specific moment in time is a conscience choice by the developer. I'm a bit torn on this. At the moment I'm more leaning towards an exception to make it explicit for the developer that they're passing an incorrect date.

There's also the case of eventually consistency here. If the developer sets a moment in the future that's reached before the API call is made then it's debatable wether it should throw an exception or be handled gracefully. But I don't think we should worry about that case as it's an edge case at best.

I'm going to let @taylorotwell decide on this one.

In any case, neither is a breaking change since Stripe was already throwing an API exception.

benjamindoe:
There are instances that Cashier handles stuff for you, such as creating a Stripe customer if one doesn't already exist on the user when creating subscriptions. My current use case simply does something similar to the following. This PR simply just moved the `if()` statement to the package.

```php
if ($trial->isFuture()) {
    $this->trialUntil($trial);
}
```

Of course if it's decided to throw an exception, we now have two options; throw an internal Cashier exception so requests don't get sent to Stripe or just use Stripe's existing API exception.

driesvints:
I think if we did decide to go for the exception throwing we can roll our own. That'll prevent the API request from being sent in the first place.

taylorotwell:
I don't think we need to do anything. Just let Stripe error.

PR #1094, opened 2021-03-12
[12.x] Update payment page

driesvints:
This PR does a couple of things:

- Update Tailwind to v2
- Some minor styling tweaks
- Adds `requires_confirmation` as a valid payment type to confirm payments on this page

Closes https://github.com/laravel/cashier-stripe/issues/1092

### DIFF
--- a/resources/views/payment.blade.php
+++ b/resources/views/payment.blade.php
@@ -6,16 +6,16 @@
 
     <title>{{ __('Payment Confirmation') }} - {{ config('app.name', 'Laravel') }}</title>
 
-    <link href="https://unpkg.com/tailwindcss@^1.0/dist/tailwind.min.css" rel="stylesheet">
+    <link href="https://unpkg.com/tailwindcss@^2/dist/tailwind.min.css" rel="stylesheet">
 
     <script src="https://cdn.jsdelivr.net/npm/vue@2.6.10/dist/vue.min.js"></script>
     <script src="https://js.stripe.com/v3"></script>
 </head>
-<body class="font-sans text-gray-600 bg-gray-200 leading-normal p-4 h-full">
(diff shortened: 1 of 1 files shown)


resources/views/payment.blade.php
```diff
@@ -33,27 +33,27 @@
...
+                    <h1 class="text-xl mb-4 text-gray-600">
                         {{ __('Payment Cancelled') }}
                     </h1>
 
                     <p class="mb-6">{{ __('This payment was cancelled.') }}</p>
                 @else
                     <div id="payment-elements" v-if="! paymentProcessed">
                         <!-- Payment Method Form -->
-                        <div v-show="requiresPaymentMethod">
+                        <div v-show="requiresPaymentMethod || requiresConfirmation">
```
benjamindoe, line 54: I think this is the wrong place for this check, it's much more like `requiresAction` as it doesn't require the user to enter payment method details.

https://github.com/laravel/cashier-stripe/blob/b21e2d625d9fa83b5fe847b9567420c466b20620/resources/views/payment.blade.php#L97

benjamindoe: See also:

https://stripe.com/docs/payments/intents

benjamindoe (changes requested):
I think this is the wrong place for this check, it's much more like `requiresAction` as it doesn't require the user to enter payment method details.

https://github.com/laravel/cashier-stripe/blob/b21e2d625d9fa83b5fe847b9567420c466b20620/resources/views/payment.blade.php#L97

driesvints:
@benjamindoe good catch. Don't know why I didn't do that. I remember thinking the same a week ago. Thanks!

benjamindoe (approved):

PR #1096, opened 2021-03-13
[13.x] Cashier Stripe Factories

### LINKED ISSUE #1072: Implement factories
Implement Laravel v8 factories for the Subscription and SubscriptionItem models. Needs to go to master.

Load them in the same way as Passport: https://github.com/laravel/passport/blob/10.x/composer.json#L43

sonja-turo:
This is a factory implementation for Subscription and SubscriptionItem. Please be gentle. This is my first submission to any open source project (but I've been developing in Laravel for years).

The Subscription Factory contains helper state functions such as incomplete() to easily set the status without having to know the Stripe constants (especially if they change). The SubscriptionItem contains no helpers as I couldn't really find any specific state on the item.

The Cashier SubscriptionTest unit test has been modified to extend the Cashier TestCase rather than the base PhpUnit TestCase due to the factory needing access to the config function, so that the factory will work against custom parent models of the subscription defined in app('cashier.model'). All tests in the SubscriptionTest have been modified to use the factory and the hlper classes.

It has been tested in a standard Laravel application against the standard User and custom models.

Closes #1072

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,8 @@
     },
     "autoload": {
         "psr-4": {
-            "Laravel\\Cashier\\": "src/"
+            "Laravel\\Cashier\\": "src/",
+            "Laravel\\Cashier\\Database\\Factories\\": "database/factories/"
         }
     },
     "autoload-dev": {
--- a/database/factories/SubscriptionFactory.php
+++ b/database/factories/SubscriptionFactory.php
@@ -28,17 +28,17 @@ class SubscriptionFactory extends Factory
         return [
             (new $model)->getForeignKey() => ($model)::factory(),
             'name' => $this->faker->title,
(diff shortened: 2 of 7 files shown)


driesvints (approved):

driesvints:
Thanks @sonja-turo! :)

driesvints:
I btw removed the changes from the unit tests because it's intended that they rely on PHPUnit test case class so the framework isn't booted.

sonja-turo:
Thanks for the feedback. I was a little iffy on it, but I did see other unit tests that boot the framework, so I assumed this was an ok thing since the tests still passed. Any other feedback on the pull aside from the styling which I've taken note of?

driesvints:
It's true that the webhooks test also use the TestCase class but they don't make any stripe api calls so they're still very fast to run. Nope, that's it.

PR #1113, opened 2021-04-06
[11.x] Replace super heavy symfony/intl dependency

GrahamCampbell:
`symfony/intl` is 16MB and is not even used by this package code or any of its dependencies.


driesvints:
11.x isn't maintained anymore

PR #1114, opened 2021-04-07
[12.x] Replace super heavy symfony/intl dependency

GrahamCampbell:
`symfony/intl` is 16MB and is not even used by this package code or any of its dependencies. In 13.x, we should probably remove this entirely, since `ext-intl` is already suggested.


driesvints:
I'll take a look at this later this week.

driesvints (approved):
This lgtm. @GrahamCampbell we're not going to remove this because we don't want to make `ext-intl` a hard requirement for Cashier.

PR #1129, opened 2021-04-15
[13.x] Fix trialEndsAt more closely represents onTrial

### LINKED ISSUE #1128: Unexpected behaviour with onTrial method in Billable
- Cashier Version: 12.9.1
- Laravel Version: 7.30.4
- PHP Version: 7.4.16

### Description:

Not so much a bug as unexpected behaviour from the onTrial method.

I have a snippet of code that does something similar to the following:

```php
if ($user->onTrial()) {
    $trialEnd = $user->trialEndsAt()->format('d/m/y');
}
```
The behaviour I would expect is that the `onTrial` and `trialEndsAt` should both check and return the same date & time respectively. The issue in question arises when a user is on a generic trial and has a subscription that isn't on a trial.

`onTrial` first checks for generic trial *if* no arguments have been passed through. It then defaults to the subscription named `default` and checks that subscription for a trial.

`trialEndsAt` first checks the `default` subscription if there are no arguments passed through. If a trial doesn't exist there it falls back to generic trial.

The point I am trying to get across is that both `onTrial` and `trialEndsAt` should probably follow similar logic when it comes to checking for trials. Perhaps `onTrial` should call `trialEndsAt`?

benjamindoe:
Closes #1128 

The `trialEndsAt` returns the `trial_ends_at` attribute by default if there are no arguments passed through and the user is indeed on a generic trial. Without the `onGenericTrial` check, the function will always return `trial_ends_at` if no arguments were passed through.

The `trialEndsAt` function also returns `trial_ends_at` as a fallback because there may be a use case that an application expects a historical date from `trialEndsAt`.

### DIFF
--- a/src/Concerns/ManagesSubscriptions.php
+++ b/src/Concerns/ManagesSubscriptions.php
@@ -60,6 +60,10 @@ trait ManagesSubscriptions
      */
     public function trialEndsAt($name = 'default')
     {
+        if (func_num_args() === 0 && $this->onGenericTrial()) {
+            return $this->trial_ends_at;
+        }
+
         if ($subscription = $this->subscription($name)) {
             return $subscription->trial_ends_at;
         }
--- a/tests/Feature/SubscriptionsTest.php
+++ b/tests/Feature/SubscriptionsTest.php
@@ -488,8 +488,11 @@ class SubscriptionsTest extends FeatureTestCase
         $user->newSubscription('main', static::$planId)
             ->create('pm_card_visa');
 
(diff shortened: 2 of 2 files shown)


driesvints:
@benjamindoe Can you confirm this doesn't breaks other tests?

benjamindoe:
All tests pass locally. Tested on 8 and 7.4.

driesvints (approved):

driesvints:
Thanks @benjamindoe

Issue #1071, opened 2021-02-22
Support for creating Tax IDs on the customer


benjamindoe:
Currently, there's no way of creating Customer Tax IDs from the user object. This must be done using the Stripe SDK directly
```php
\Stripe\Customer::createTaxId($user->stripe_id, [
    'type' => \Stripe\TaxId::TYPE_EU_VAT,
    'value' => 'DE12345678'
], $user->stripeOptions());
```

Currently, you can get all Tax IDs using the Stripe customer directly, (which isn't so bad)
```php
$user->asStripeCustomer()->tax_ids
```

It would be nice to have full integration with tax IDs to make things a bit more fluent and Eloquent relationship-like, for example

```php
$user->taxIds()->create([
    'type' => 'eu_vat', // Possibly throw Exception if unknown ENUM?
    'value' => 'DE12345678'
]);
```
If that's a bit much, simple helpers to the Stripe SDK would also be nice
```php
$user->taxIds(); // Return instance of \Illuminate\Support\Collection
$user->taxId('txi_IzcmyhZukd4oKD'); // Retrieve Tax ID by Stripe ID
$user->createTaxId($type, $value);
$user->deleteTaxId('txi_IzcmyhZukd4oKD'); // delete Tax ID by Stripe ID

```

There's no fluent way of using the Stripe customer object directly as the `createTaxId` function on `\Stripe\Customer` is static, which is unfortunate.

driesvints:
Thanks @benjamindoe. We'll be looking at adding some nice API's around Tax Id's soon as well as customer billing address details. Appreciating PR's however in the meantime. I like your proposal of making a fluent API but without making that actually Eloquent like of course.

driesvints:
Here's the PR: https://github.com/laravel/cashier-stripe/pull/1137

PR #1137, opened 2021-04-23
[12.x] Add methods for managing TaxIDs

driesvints:
This PR adds some methods to easily manage TaxID's on a customer. This is useful for keeping this data in Stripe so it also appears on its invoices an later in our own receipts like in https://github.com/laravel/cashier-stripe/pull/1136

Usage:

```php
// Return all of the customer's TaxIDs in an collection instance...
$user->taxIds(); 

// Retrieve a specific TaxID from the customer...
$user->findTaxId('txi_IzcmyhZukd4oKD');

// Create a new TaxID for the customer...
$user->createTaxId($type, $value);

// Delete a TaxID from the customer...
$user->deleteTaxId('txi_IzcmyhZukd4oKD');
```

Closes https://github.com/laravel/cashier-stripe/issues/1071

### DIFF
--- a/src/Concerns/ManagesInvoices.php
+++ b/src/Concerns/ManagesInvoices.php
@@ -131,7 +131,7 @@ trait ManagesInvoices
             $stripeInvoice = StripeInvoice::retrieve(
                 $id, $this->stripeOptions()
             );
-        } catch (Exception $exception) {
+        } catch (StripeInvalidRequestException $exception) {
             //
         }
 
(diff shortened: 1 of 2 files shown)


src/Concerns/ManagesInvoices.php
```diff
@@ -131,7 +131,7 @@ public function findInvoice($id)
             $stripeInvoice = StripeInvoice::retrieve(
                 $id, $this->stripeOptions()
             );
-        } catch (Exception $exception) {
+        } catch (StripeInvalidRequestException $exception) {
```
driesvints, line 134: This was apparently referencing an invalid exception. Seems like no-one has hit this snag so far...

PR #1146, opened 2021-05-13
[13.x] Preview upcoming invoice

### LINKED ISSUE #1142: Improve upcoming invoices
Allow to pass data to upcoming invoices so subscription changes can be shown, etc. Maybe add a method to the subscription itself.

Look into https://stripe.com/docs/api/invoices/upcoming

benjamindoe:
Noticed this issue and thought I would suggest a PR since I've done something similar in my current project. Nothing fancy. Just a very basic method that has the same signature as the `swap` method

```php
$upcomingInvoice = $subscription->previewInvoice('price_to_swap')
```

Closes #1142

### DIFF
--- a/src/Concerns/ManagesInvoices.php
+++ b/src/Concerns/ManagesInvoices.php
@@ -108,7 +108,9 @@ trait ManagesInvoices
         }
 
         try {
-            $stripeInvoice = StripeInvoice::upcoming(['customer' => $this->stripe_id], $this->stripeOptions());
+            $stripeInvoice = StripeInvoice::upcoming(array_merge([
+                'customer' => $this->stripe_id,
+            ], $options), $this->stripeOptions());
 
             return new Invoice($this, $stripeInvoice);
         } catch (StripeInvalidRequestException $exception) {
--- a/src/Concerns/ManagesInvoices.php
+++ b/src/Concerns/ManagesInvoices.php
@@ -101,7 +101,7 @@ trait ManagesInvoices
      * @param  array  $options
      * @return \Laravel\Cashier\Invoice|null
      */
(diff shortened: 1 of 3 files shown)


driesvints:
Cool stuff, thanks @benjamindoe! I think we can improve this further. Can we re-use the `upcomingInvoice` method from the billable? I also think we need a separate general `upcomingInvoice` method on the subscription itself that `previewUpcomingInvoice` calls. 

We can add an argument to `upcomingInvoice` [here](https://github.com/laravel/cashier-stripe/blob/532089a487dd09c6fa5adaf9c0088715a7623e48/src/Concerns/ManagesInvoices.php#L105) so we can pass options and don't need to re-use the `stripeOptions` call etc.

Also: I believe we need to re-use `swapOptions` as well, no?

benjamindoe:
For sure, I kind of just craved this out of my current app without much refactoring.

I'll look into reusing `swapOptions`. There must have been a reason why I didn't do it. Probably the payload data is slightly different when interacting with invoices directly?

I'll look at factoring out `upcomingInvoice` and `previewUpcomingInvoice`

benjamindoe:
So unfortunately, `getSwapOptions` won't work in its current state with previewing upcoming invoices. The payload is incompatible due to calling Stripe invoices directly instead of via a subscription. A good example of this is `subscription_items` for invoices vs. `items` for subscription payloads.

driesvints:
Okay. Thanks for working on this @benjamindoe. I'll take it from here.

driesvints:
Thanks @benjamindoe. I've finished this PR. Managed to re-use `swapOption` in a different way.

benjamindoe:
Nice one... Interesting method

Issue #1145, opened 2021-05-11
Multiple discounts


driesvints:
Stripe added support for multiple discounts. I've already made progress on a PR and hope to land it in v13.

PR #1147, opened 2021-05-13
[13.x] Multiple discounts on receipts

driesvints:
This PR implements multiple discount support for receipts. With these changes, all discounts on a receipt are properly displayed with their name set up through Stripe as well as their percentage or amount off.

Here's an example of multiple discounts:

Screenshot 2021-05-14 at 18 49 09

The general amount of is calculated first. Afterwards, individual discounts are applied per specific product they apply for.

Here's the equivalent Stripe receipt:

Screenshot 2021-05-14 at 18 51 01

Some other changes in this PR:

- I've refactored the way extra parameters on the Stripe Invoice object are expanded. They are now only expanded when called upon and the data is cached internally on the object.
- A new `Discount` and `Coupon` object were introduced to wrap the related Stripe objects.
- Cleaned up some inconsistencies with collection class instantiation and `is_null` calls.

Closes https://github.com/laravel/cashier-stripe/issues/1145

### DIFF
--- a/resources/views/receipt.blade.php
+++ b/resources/views/receipt.blade.php
@@ -246,6 +246,7 @@
                         </tr>
                     @endif
 
+
                     <!-- Display The Discount -->
                     @if ($invoice->hasDiscount())
                         <tr>
--- a/resources/views/receipt.blade.php
+++ b/resources/views/receipt.blade.php
@@ -30,7 +30,7 @@
             vertical-align: bottom;
             font-weight: bold;
             padding: 8px;
-            line-height: 20px;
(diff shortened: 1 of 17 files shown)


resources/views/receipt.blade.php
```diff
@@ -41,7 +41,7 @@
 
         .table td {
             padding: 8px;
-            line-height: 20px;
+            line-height: 14px;
```
driesvints, line 44: I decreased the line-height here slightly to account for the extra items on the invoice.

Issue #1153, opened 2021-05-17
Charge with Invoice quantity bug


ErnestStaug:
- Cashier Version: 12.13.1
- Laravel Version: 8.41.0
- PHP Version: 8
- Database Driver & Version: Mysql 8

### Description:
Cannot  single charge user with amount and quantity in same request.

I get error:
`You may only specify one of these parameters: amount, quantity.`

If `quantity` is specified there should be `unit_amount` as per stripe [documentation](https://stripe.com/docs/api/invoiceitems/create#create_invoiceitem-unit_amount). Therefore both things cannot exist in same array when we create "Tab"/"StripeInvoiceItem"

### Steps To Reproduce:

Just try second example in Laravel [documentation](https://laravel.com/docs/8.x/billing#charge-with-invoice)

driesvints:
I'm checking into this with Stripe. These parameters are super confusing to me.

driesvints:
I've sent in a fix for this: https://github.com/laravel/cashier-stripe/pull/1161

PR #1161, opened 2021-05-20
[12.x] Fix adding invoice item with quantities

driesvints:
Apparently if you use a quantity here you're required to use `unit_amount` instead of `amount`. No idea why no-one ever stumbled on this before.

Fixes https://github.com/laravel/cashier-stripe/issues/1153

### DIFF
--- a/src/Concerns/ManagesInvoices.php
+++ b/src/Concerns/ManagesInvoices.php
@@ -30,10 +30,15 @@ trait ManagesInvoices
 
         $options = array_merge([
             'customer' => $this->stripe_id,
-            'amount' => $amount,
             'currency' => $this->preferredCurrency(),
             'description' => $description,
         ], $options);
+
+        if (array_key_exists('quantity', $options)) {
+            $options['unit_amount'] = $amount;
+        } else {
+            $options['amount'] = $amount;
+        }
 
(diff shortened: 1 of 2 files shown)


PHLAK:
FYI, this (minor) change broke our app after a routine dependency update. Before this change we were doing the following

```php
$account->tab($description, null, [
    'unit_amount' => $unitAmount,
    'quantity' => $quantity,
]);
```

This resulted in `amount` being `null` and `unit_amount` and `quantity` being set in the request as expected.

However, after this change `unit_amount` was being overridden with the value of `amount` (`null`) and resulted in neither `amount` nor `unit_amount` being set in the request. That in turn resulted in the same error mentioned in #1153 that this PR was supposed to fix.

I'm not saying this PR is bad or wrong, and we may have been misusing this method to begin with. I just wanted to shed some light in case someone else runs into the same issue.

driesvints:
@PHLAK thanks for reporting that. Definitely a use case I wasn't thinking of. I tagged v12.14.1 with a fix for that.

You could just upgrade to the following now:

```php
$account->tab($description, $unitAmount, [
    'quantity' => $quantity,
]);
```

PHLAK:
Thanks @driesvints, this is what we're doing now.

Issue #1163, opened 2021-05-21
Stripe Checkout not working with Coupons


io238:
- Cashier Version: 12.13.0
- Laravel Version: 8.40.0
- PHP Version: 8.0.3

### Description:
It's not possible to generate a Stripe Checkout session through Cashier when using Coupons via the `->withCoupon()` method.

#### There are two issues:
(1) The Stripe API does not accept both `allow_promotion_codes` and `discounts` parameters. Stripe throws an exception even when the `allow_promotion_codes` parameter is set to `false`.

(2) The `discount` parameter in the session creation payload needs to be an indexed array.

Both issues seem to be in inconsistent implementation of the Checkout API. It's not very well documented and this different from the way how regular subscriptions are created (without Checkout). See https://stripe.com/docs/billing/subscriptions/discounts for more details.

### Steps To Reproduce:
```
$checkoutSession = Auth::user()
  ->newSubscription('prod_xxxxxx', 'price_xxxxxx')
  ->withCoupon('xxxxxx')
  ->checkout();
```

#### API throws the following exception:
```
You may only specify one of these parameters: allow_promotion_codes, discounts. 
```

### Suggested solution:
When using a Coupon, the `allow_promotion_codes` needs to be removed from the options instead of passing it along with the value `false`. Also, the discount needs to be passed as a single-element indexed array.

I'd be happy to create a PR for this.

PR #1164, opened 2021-05-21
Bugfix: Creating Checkout session with Coupon (#1163)

io238:
This PR fixes issue #1163.

### Description:
It's not possible to generate a Stripe Checkout session through Cashier when using Coupons via the `->withCoupon()` method.

#### There are two issues:
(1) The Stripe API does not accept both `allow_promotion_codes` and `discounts` parameters. Stripe throws an exception even when the `allow_promotion_codes` parameter is set to `false`.

(2) The `discount` parameter in the session creation payload needs to be an indexed array.

Both issues seem to be in inconsistent implementation of the Checkout API. It's not very well documented and this different from the way how regular subscriptions are created (without Checkout). See https://stripe.com/docs/billing/subscriptions/discounts for more details.

### Changes:
- Empty values such as `'allow_promotion_codes' = false` are filtered from the payload before sending it to Stripe
- `coupon` nested one level deeper in the `discounts` parameter

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -381,19 +381,21 @@ class SubscriptionBuilder
             $trialEnd = null;
         }
 
-        return Checkout::create($this->owner, array_merge([
+        return Checkout::create($this->owner, array_filter(array_merge([
             'mode' => 'subscription',
             'line_items' => collect($this->items)->values()->all(),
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @io238. I've took an alternative approach for this here: https://github.com/laravel/cashier-stripe/pull/1165

Issue #956, opened 2020-06-09
Refactor internals to products and prices


driesvints:
Stripe has rolled out their new products and prices APIs which replace the concept of subscription plans. Even though things will continue to keep on working properly, it's probably best that we rewrite the internals eventually and update the docs.

Also check for webhook events (switch from plan key to price key).

Rename stripe_plan database columns to stripe_price?

Docs: https://stripe.com/docs/billing/prices-guide
Migration guide: https://stripe.com/docs/billing/migration/migrating-prices

Multi-plan -> multi-product: https://stripe.com/docs/billing/subscriptions/multiple-products
Replace `$stripeSubscription->plan` and `$stripeSubscription->quantity` calls
Remove subscription quantity & stripe_plan columns?

driesvints:
Already did a small update for the docs: https://github.com/laravel/docs/pull/6165

Atm this issue isn't urgent since everything will keep on working as is. If we ever do a new major version again we can revisit.

sebdesign:
@driesvints I'm using products and prices for a new project, and I'm willing to implement these features. Is there any plan on how should this refactored?

I have defined a price with `metered` billing which means when I create a subscription I need to pass an empty array for the `$plans` parameter, and and manually pass a `price` instead of a `plan` without a `quantity` to the `$subscriptionOptions`.

```php
$user->newSubscription('default', [])->create($paymentMethod, [], [
    'items' => [
        [
            'price' => 'price_HGd7M3DV3IMXkC',
        ],
    ],
]);
```

Now this fails because when Cashier is creating the subscription item in the database it needs the `quantity`, so I've override the `subscription_items` migration like this:

```php
$table->integer('quantity')->nullable();
```

driesvints:
@sebdesign this PR is only as a reminder to rename some of the internals from "plans" to "prices", variable names, method arguments etc. I don't think any big changes are needed as everything already works. 

Metered billing in general isn't supported by Cashier yet and has its own separate issues here: https://github.com/laravel/cashier-stripe/issues/555

fkraefft:
> @driesvints I'm using products and prices for a new project, and I'm willing to implement these features. Is there any plan on how should this refactored?
> 
> I have defined a price with `metered` billing which means when I create a subscription I need to pass an empty array for the `$plans` parameter, and and manually pass a `price` instead of a `plan` without a `quantity` to the `$subscriptionOptions`.
> 
> ```
[...]

PR #1166, opened 2021-05-21
[13.x] Rename plans to prices

driesvints:
This PR renames all "plan" terminology to prices. Stripe has adopted the new Price object for a while now and has phased out all plan functionality. The old API's still work but we should take the effort to keep things on par with Stripe. 

I'll write a migration for this when I'm ready to work on the upgrade guide.

Closes https://github.com/laravel/cashier-stripe/issues/956


top-master:
**See** [Revert renaming Plans to Prices (feature or fix request)](https://github.com/laravel/cashier-stripe/issues/1341)

**As Web-Hooks prove:**

**1** Each time during a Stripe `Subscription` creation, if a Stripe-`Price`-object was added intto "items" instead of a Stripe-`Plan`-object, then Stripe automatically creates a `Plan` (in their DB), and that one for each `Price` added.

**2** Also, each time we create a Stripe `Plan` manually, Stripe automatically creates a Stripe `Price` (in their DB), to store things like `amount` and `currency`, and auto-deletes that `Price` later (once `Plan` is deleted).

**3** **But the *huge* difference is** that:
* A `Plan` forces us to set "`interval`" during creation (while "interval" is optional for `Price` objects), hence `Plan`-object is optimal for subscriptions.
* A `Plan` can have different `amount` then the `Product`'s `Price`.
* Let's consider we have one "newspaper" as Product, and we sell a single newspaper by a higher *`Price`* (if people have no subscription), which of below makes no sense:
    * "Our Subscription *`Plan`* has separate lower `Price`."
    * "Our Subscription *`Price`* has separate lower `Price`."

In other words, each `Subscription`-item needs both a `Plan` and a `Price`, and although Stripe may create requirements automatically, above proves it was a *huge* mistake to rename everything from `Plans` to `Prices` (IMHO).

If you want to clarify the difference of `Plans` and `Prices` to developers, then simply just do that, by writting somewhere:
>A `Plan` is the requirement to create `Subscriptions` (because it stores `interval`, `interval_count` and other settings), but if a `Price` has `recurring.interval` set, then Stripe will automatically create a `Plan` for us.

driesvints:
See my response here: https://github.com/laravel/cashier-stripe/issues/1341#issuecomment-1087190407

PR #1167, opened 2021-05-24
[12.x] Add extra methods to Invoice object

driesvints:
Introduces some extra methods for the Invoice object that can be used in Spark amongst else.


taylorotwell:
Merged.

PR #1171, opened 2021-05-27
Retrieve all type of payment methods

Yo1L:
Hi,
The default parameters can now be overridden with this fix.
In my purpose, I would like to retrieve any kind of payment type (in my case, 'sepa_debit') by defining it in the paramaters.
Thanks for your time


driesvints:
Hi, this is coming in v13. Just have to wait a little longer :)

PR #1178, opened 2021-06-03
[13.x] Allow customers to be synced with Stripe

driesvints:
This PR will allow customers to easily be synced with Stripe. The developer would wire this using the `updated` model event:

```php
/**
 * The "booted" method of the model.
 *
 * @return void
 */
protected static function booted()
{
    static::updated(queueable(function ($customer) {
        $customer->syncStripeCustomerDetails();
    }));
}
```


kyranb:
Should `syncStripeCustomerDetails` take an optional `metadata` parameter?

Currently we have a job that runs that passes to Stripe an array of metadata like:

`account_owner_id`
`tenant_id`
`tenant_created_at`

driesvints:
@kyranb the idea is that you'd overwrite this method in your billable to add whatever you want.

themsaid (approved):

PR #1198, opened 2021-06-17
[13.x] Implement new calculateTaxes call

driesvints:
This implements a new `Cashier::calculateTaxes()` static method that'll allow you enable automatic tax calculation on all relevant Stripe calls. By default it follows Stripe's behavior and is set to `false`. Additionally, the previous `withTax` method is now deprecated as it is not needed anymore. You can now use your code as regularly without the need of using extra methods. This method call goes into the boot method of your `ApServiceProvider`.

Also, the `tax_id_collection` is now enabled by default when the calculation of taxes is enabled. This PR also fixes a typo for an forgotten rename of `collectsTaxIds`.

I'll send in a PR to the docs when this one is accepted.

### DIFF
--- a/src/Concerns/PerformsCharges.php
+++ b/src/Concerns/PerformsCharges.php
@@ -85,7 +85,7 @@ trait PerformsCharges
 
                 return $item;
             })->values()->all(),
-            'tax_id_collection' => $this->collectsTaxIds,
+            'tax_id_collection' => config('stripe.taxes') ?: $this->collectTaxIds,
         ]);
 
         return Checkout::create($this, array_merge($payload, $sessionOptions), $customerOptions);
--- a/src/Concerns/PerformsCharges.php
+++ b/src/Concerns/PerformsCharges.php
@@ -3,6 +3,7 @@
 namespace Laravel\Cashier\Concerns;
 
 use Illuminate\Support\Collection;
+use Laravel\Cashier\Cashier;
 use Laravel\Cashier\Checkout;
 use Laravel\Cashier\Payment;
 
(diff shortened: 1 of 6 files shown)


src/Concerns/PerformsCharges.php
```diff
@@ -23,7 +24,6 @@ trait PerformsCharges
     public function charge($amount, $paymentMethod, array $options = [])
     {
         $options = array_merge([
-            'automatic_tax' => $this->automaticTaxPayload(),
```
driesvints, line 26: I removed this because it's not possible to use automatic tax with payment intents atm it seems.

src/Subscription.php
```diff
@@ -791,7 +789,6 @@ protected function mergeItemsThatShouldBeDeletedDuringSwap(Collection $items)
     protected function getSwapOptions(Collection $items, array $options = [])
     {
         $payload = [
-            'automatic_tax' => $this->automaticTaxPayload(),
```
driesvints, line 794: I removed this because disabling automatic taxes during a swap is unlikely to happen. People can still do this by passing options explicitly if they want.

src/Subscription.php
```diff
@@ -1132,7 +1129,6 @@ public function latestInvoice()
     public function upcomingInvoice(array $options = [])
     {
         return $this->owner->upcomingInvoice(array_merge([
-            'automatic_tax' => $this->automaticTaxPayload(),
```
driesvints, line 1135: I removed this because the underlying `upcomingInvoice` already adds this.

PR #1214, opened 2021-07-02
[13.x] Switch to Illuminate\Http\Response

chasenyc:
Switching the WebhookController to use Laravel's standard http response. This came up for me after adding a new middleware based on [recent changes](https://github.com/laravel/framework/pull/37847) in Laravel 8.49:

```php
    public function handle($request, Closure $next)
    {
        $requestId = (string) Str::uuid();

        Log::withContext([
            'request-id' => $requestId
        ]);

        return $next($request)->header('Request-Id', $requestId);
    }
```

I noticed in my QA environment that all of my webhooks were failing because of the following error: 
```
Call to undefined method Symfony\\Component\\HttpFoundation\\Response::header() at /home/forge/default/app/Http/Middleware/AssignRequestId.php:26
```

I'm not sure if there is a historical or current reason to opt for Symfony response over Illuminate response which extends Symfony response.

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -3,6 +3,7 @@
 namespace Laravel\Cashier\Http\Controllers;
 
 use Illuminate\Http\Request;
+use Illuminate\Http\Response;
 use Illuminate\Notifications\Notifiable;
 use Illuminate\Routing\Controller;
 use Illuminate\Support\Carbon;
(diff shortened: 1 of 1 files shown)


driesvints:
Technically this will be a breaking change because people are currently returning a Symfony response if they overwrite the method in their own apps. I don't immediately see this being a big issue. Most likely things will continue to work as before and it does make sense to use the Illuminate one instead of the Symfony one. I also don't immediately know what the historical reasons are for choosing the Symfony one. Probably @taylorotwell knows.

chasenyc:
Sorry, I am not too experienced with this but how does one go about make a PR for a breaking change, i.e. I'm guessing 14.x?

driesvints:
Yes (to master) but let's see what @taylorotwell says

taylorotwell:
If it's a breaking change we can't merge it.

driesvints:
@chasenyc honestly, I'd just leave this one be. You can always wrap it in an illuminate response if you want to.

PR #1215, opened 2021-07-02
```stripe_id``` collation set to ```utf8_bin```

chinmaypurav:
[Official Stripe Documentation for collation](https://stripe.com/docs/upgrades#what-changes-does-stripe-consider-to-be-backwards-compatible)


driesvints:
This will break PostgreSQL apps. See https://github.com/laravel/cashier-stripe/issues/747

There's already a note in the docs about this:

Screen Shot 2021-07-02 at 11 12 16

PR #1218, opened 2021-07-08
[13.x] Server side Checkout redirect

driesvints:
This PR implements some nice API's to Stripe's recent addition of exposing the Checkout url on its session object. This will greatly simplify the workflow for a library like Cashier to use redirects instead of doing it through Stripe.js

Before:

```php
$checkoutBtn = Auth::user()->checkout('price_tshirt')->button();

return view('checkout', compact('checkoutBtn'));
```

And have the customer click a button. Now:

```php
return Auth::user()->checkout('price_tshirt');
```

I've deprecated the `button` method because I think that from now on that's going to be a very rare use case.

### DIFF
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -4,11 +4,13 @@ namespace Laravel\Cashier;
 
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Jsonable;
+use Illuminate\Contracts\Support\Responsable;
+use Illuminate\Support\Facades\Redirect;
 use Illuminate\Support\Facades\View;
 use JsonSerializable;
 use Stripe\Checkout\Session;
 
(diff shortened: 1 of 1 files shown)


src/Checkout.php
```diff
@@ -76,6 +79,27 @@ public function button($label = 'Check out', array $options = [])
...
     }
 
+    /**
+     * Redirect to the checkout session.
+     *
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function redirect()
+    {
+        return Redirect::to($this->session->url, 303);
```
driesvints, line 89: I chose 303 on purpose here because Stripe recommends it in their docs (and @cjavilla-stripe did in one of his video tutorials): https://stripe.com/docs/payments/accept-a-payment?platform=web&ui=checkout#redirect-customers

PR #1273, opened 2021-11-04
Allow the base url for the Stripe API to be customised

seanmtaylor:
This PR allows the base url of the Stripe API to be changed if needed, such as when using https://github.com/stripe/stripe-mock.

If there's already a way to change the base url without needing this change then please let me know what I'm missing as I couldn't see a way to customise the `$options` parameter of the `stripe()` method.

### DIFF
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -7,6 +7,7 @@ use Money\Currency;
 use Money\Formatter\IntlMoneyFormatter;
 use Money\Money;
 use NumberFormatter;
+use Stripe\BaseStripeClient;
 use Stripe\Customer as StripeCustomer;
 use Stripe\StripeClient;
 
@@ -25,6 +26,13 @@ class Cashier
      * @var string
      */
     const STRIPE_VERSION = '2020-08-27';
+
+    /**
+     * The base url for the Stripe API
+     *
+     * @var string
+     */
+    public static $apiBaseUrl = BaseStripeClient::DEFAULT_API_BASE;
 
     /**
      * The custom currency formatter.
@@ -106,6 +114,7 @@ class Cashier
         return new StripeClient(array_merge([
             'api_key' => $options['api_key'] ?? config('cashier.secret'),
             'stripe_version' => static::STRIPE_VERSION,
+            'api_base' => static::$apiBaseUrl,
         ], $options));
     }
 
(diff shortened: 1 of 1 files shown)


driesvints:
I'm not sure what you're trying to achieve. This should already be the default. Why define it explicitly? 

Can you show where you need to adjust the url with https://github.com/stripe/stripe-mock ?

driesvints:
Thanks @seanmtaylor. I think this change can be added but I want to warn you about Stripe Mock: since it's stateless I suspect a lot of this library won't be compatible again. Internally, Cashier does a lot of synchronous API calls to Stripe that rely on the data to be persistent for subsequent calls. Therefor something like Stripe Mock isn't useful here. I tried it out in our test suite and the tests failed for the above reasons. 

I'd much more want to have a look at https://github.com/stripe/stripe-playback-node but there doesn't seems to be a PHP implementation yet.

In the meantime though, I don't see a specific reason to not add this change. But I hope I informed you enough about the limitations of Stripe Mock.

driesvints (approved):

seanmtaylor:
> Thanks @seanmtaylor. I think this change can be added but I want to warn you about Stripe Mock: since it's stateless I suspect a lot of this library won't be compatible again. Internally, Cashier does a lot of synchronous API calls to Stripe that rely on the data to be persistent for subsequent calls. Therefor something like Stripe Mock isn't useful here. I tried it out in our test suite and the tests failed for the above reasons.

Thanks @driesvints, much appreciated. I'm aware of the limitations, I saw them mentioned in one of your blog posts (https://driesvints.com/blog/testing-cashier/). 

My thoughts are at least for the simple calls which don't rely on a persistent state, then I can use stripe-mock, for others that's where I can implement a more advanced solution, like those outlined in your blog post.

> 
> I'd much more want to have a look at https://github.com/stripe/stripe-playback-node but there doesn't seems to be a PHP implementation yet.

That link doesn't work for me, I wonder if it's a private repo.

Just done a bit of googling to find out more and it seems like to make use of that, it will require changing the url to point at the running instance of stripe-playback. So this PR should hopefully come in handy for that too :crossed_fingers:

driesvints:
@seanmtaylor the repo I linked to is indeed private, sorry.

PR #1280, opened 2021-11-19
[13.x] Allow latest moneyphp version

mmachatschek:
This is a breaking change for some users, as moneyphp version 4 requires the bcmath extension

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          extensions: dom, curl, libxml, mbstring, zip
+          extensions: dom, curl, libxml, mbstring, zip, bcmath
           tools: composer:v2
           coverage: none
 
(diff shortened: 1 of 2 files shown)


driesvints:
Hmm, technically we could add this to the current stable release. Composer will handle the extension checking. @taylorotwell what do you think?

driesvints:
Almost every Forge site should have bcmath installed and did an ask around for people using Cashier/Spark Stripe. All had bcmath installed. So I think we can move this to 13.x.

mmachatschek:
I rebased the PR to the 13.x branch

driesvints (approved):

taylorotwell:
I dunno - is it not a breaking change to now require a new PHP extension? I don't see how that is not a breaking change.

driesvints:
@taylorotwell adding new dependencies without breaking the public API is allowed under SemVer: https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

Composer will prevent Cashier from being installed onto systems which don't have bcmath bundled in PHP. I believe the impact of this will be very minimal. The only thing people need to do is compile bcmath with PHP onto their systems if they want to composer update to the latest Cashier.

Issue #1278, opened 2021-11-17
Discrepancy between factory and query


jasonvarga:
- Cashier Version: 13.6.0
- Laravel Version: 8.68.1
- PHP Version: 7.4

### Description:

This isn't really a bug (yet?) but maybe just a question.

I was adjusting some of my tests and app code to use the new factories and thought I could leverage the existing Cashier factories.

I noticed that the `SubscriptionFactory`'s `canceled` state just sets the `stripe_status`

https://github.com/laravel/cashier-stripe/blob/4bd3b2488cb107b5d5160f97afbc2960a28a1393/database/factories/SubscriptionFactory.php#L86-L91

But when you query for cancelled models or use the cancelled method itself, it uses the `ends_at` column.

https://github.com/laravel/cashier-stripe/blob/4bd3b2488cb107b5d5160f97afbc2960a28a1393/src/Subscription.php#L296-L299
https://github.com/laravel/cashier-stripe/blob/4bd3b2488cb107b5d5160f97afbc2960a28a1393/src/Subscription.php#L307-L310

Is there a reason this is different?

### Steps To Reproduce:

Using this quick factory creation and assertion, I'd expect it to be true.

```php
$sub = Subscription::factory()->canceled()->create();

$this->assertTrue($sub->cancelled()); // it's false.
```

### Additional Details

Also, and it's less of an issue, but you have `canceled` with one L in the factory and `cancelled` with 2 Ls in the model. Stripe uses `canceled`.

jasonvarga:
I'm sure you know this, but I think `$subscription->cancelled()` only looks at `ends_at` for backwards compatibility for subscriptions created before the `stripe_status` column was introduced.

driesvints:
> Is there a reason this is different?

No, this is an oversight from https://github.com/laravel/cashier-stripe/pull/1096 when they were first introduced. "cancelled" in Cashier means when an ending period has been set (even if it's in the future and not yet actually cancelled). This is to give time for a "grace period" in which the customer can still use the remaining active time. You can read more about this here: https://laravel.com/docs/8.x/billing#cancelling-subscriptions

It's not the same as "canceled" in Stripe. A "canceled" status for Stripe only is set when the subscription is actually cancelled. This is because Stripe does not consider the subscription to be cancelled when it's in its grace period. "$subscription->cancelled()" in Cashier will return `true` even when it's in its grace period. This is to determine if in fact, the subscription is cancelled or will be cancelled in the future.

[...]

PR #1281, opened 2021-11-19
[14.x] Rename cancelled to canceled

driesvints:
This PR contains breaking changes but I believe it's a sensible to make. Stripe maintains the American "canceled" while we in Cashier use the British "cancelled". I believe bringing this on-par with Stripe will be easier to make the translation between Cashier/Stripe.

Breaking changes are not always wanted and there's the question if this actually benefits end users. I believe it does as it prevents confustion like in #1278 and other situations I've seen over the past years. Besides, the adjustments are very easy to make.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -586,7 +586,7 @@
 - Add ability to ignore cashier routes ([#763](https://github.com/laravel/cashier-stripe/pull/763))
 
 ### Fixed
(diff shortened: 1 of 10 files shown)


taylorotwell:
I would just offer both methods on the backend and alias one to the other so there is no breaking change - then only document one.

PR #1297, opened 2022-01-05
Fix #1296

### LINKED ISSUE #1296: Method subscription() does not retrieve the most recent subscription as expected
- Cashier Version: 13.6.1
- Laravel Version: 8.76.2
- PHP Version: 8.0.14
- Database Driver & Version: MySQL 5.7.18

### Description:
I have a user who has two subscriptions. The oldest one has been canceled and the most recent is active. 

id | user_id | name | stripe_id | stripe_price | quantity | stripe_status | trial_ends_at | ends_at | created_at | updated_at |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
100 | 534 | advisor | sub_***BIloe | price_***AbhD8 | 1 | canceled | NULL | 2022-01-05 19:38:22 | 2022-01-05 19:32:20 | 2022-01-05 19:38:22
101 | 534 | advisor | sub_***HTZrqh | price_***AbhD8 | 1 | active | NULL | NULL | 2022-01-05 19:56:42 | 2022-01-05 19:56:42

The documentation says that the most recent subscription should be returned when using the `subscription()` method on the user.

> If a user has two subscriptions with the same name, the most recent subscription will always be returned by the subscription method. For example, a user might have two subscription records named default; however, one of the subscriptions may be an old, expired subscription, while the other is the current, active subscription. The most recent subscription will always be returned while older subscriptions are kept in the database for historical review.

But when I try to retrieve the user's subscription with `$user->subscription('advisor')` I only get the oldest one with the id `100`.

### Steps To Reproduce:

1. Subscribe a user to a subscription plan
2. Cancel the subscription
3. Subscribe again to a new subscription plan
4. Retrieve the current subscription with `$user->subscription('name')`

marcbelletre:
The `subscription()` method does not behave as expected [as per the documentation](https://laravel.com/docs/8.x/billing#checking-subscription-status).

> If a user has two subscriptions with the same name, **the most recent subscription will always be returned by the subscription method**. For example, a user might have two subscription records named default; however, one of the subscriptions may be an old, expired subscription, while the other is the current, active subscription. The most recent subscription will always be returned while older subscriptions are kept in the database for historical review.

I fixed it by returning the last subscription instead of the first one.

### DIFF
--- a/src/Concerns/ManagesSubscriptions.php
+++ b/src/Concerns/ManagesSubscriptions.php
@@ -97,7 +97,7 @@ trait ManagesSubscriptions
      */
     public function subscription($name = 'default')
     {
-        return $this->subscriptions->where('name', $name)->first();
+        return $this->subscriptions->where('name', $name)->last();
     }
 
     /**


taylorotwell:
The relationship does return the lastest one first though?

```php
    /**
     * Get all of the subscriptions for the Stripe model.
     *
     * @return \Illuminate\Database\Eloquent\Relations\HasMany
     */
    public function subscriptions()
    {
        return $this->hasMany(Cashier::$subscriptionModel, $this->getForeignKey())->orderBy('created_at', 'desc');
    }
```

marcbelletre:
Well, my bad. I had overwritten the relationship in my Billable model and did not set the order there. Sorry I should have paid more attention.  

Thanks!

PR #1303, opened 2022-01-19
Fix email check on receipt

matthieumota:
Hi, a very simple fix.


driesvints:
This is intentionally. There's always a customer email. If you look above we show the customer name if it's available, otherwise an email address. The line of code you edited is a fallback for the email address when there's a customer name.

matthieumota:
Sorry, I see it but I should have think

Issue #1311, opened 2022-01-31
Don't require `dompdf/dompdf`


PHLAK:
We use `spatie/browsershot` for our app's interal PDF rendering and with the introduction of #1304 we plan to switch Cashier over to use browsershot as well (instead of `dompdf/dompdf`). However, since `dompdf/dompdf` is still a required package it will still be installed in all environments, even when unused. It would be nice if `dompdf/dompdf` were no longer a required package to eliminate an unnecessary dependency. However I know simply moving it to `suggest` isn't straight forward and thus opened this issue to discuss if you think this would be possible and how to go about making that change. If this is possible/desirable I am willing to work on a PR.

driesvints:
I agree. Attempted a PR here for the next major release since it's a breaking change: https://github.com/laravel/cashier-stripe/pull/1312

PR #1312, opened 2022-02-01
[14.x] Make dompdf optional

driesvints:
This makes dompdf an optional dependency. Now that we have decoupled it behind an interface it can be an optional dependency. And if people want to use their own renderers they can choose not to install it: https://github.com/laravel/cashier-stripe/issues/1311


PHLAK:
Can this be back ported to the `13.x` branch as well?

Issue #1315, opened 2022-02-05
Subscription validity: cancelled during free trial period


kpolicar:
- Cashier Version: v13.8.1
- Laravel Version: 8
- PHP Version: 8
- Database Driver & Version: Irrelevant

### Description:

If the Stripe subscription is canceled during the free trial period, the methods `onTrial()` and consequently `valid()` return **true** on Cashier's subscription model. This also means the method `subscribed()` on the user model (through the ManagesSubscription trait) returns **true**.

The expected behavior is that these methods should return **false**.

**Example values, assuming the current date is 2022-02-05:**
| stripe_status | ends_at | trial_ends_at |
| ----------- | ----------- | ----------- |
| cancelled | 2022-02-04 | 2022-02-06 |

### Steps To Reproduce:

1. Subscribe a user to a subscription plan with a free trial
2. Cancel the subscription through Stripe's dashboard
3. Check the user's subscription status $user->subscribed()

kpolicar:
Previously mentioned in https://github.com/laravel/cashier-stripe/issues/1306

driesvints:
This is expected. There's a grace period to take into account. If you want to cancel and end don't let the customer make use of the grace period you need to use `cancelNow`. See https://laravel.com/docs/8.x/billing#cancelling-subscriptions

Canceling through the Stripe dashboard will always invoke the grace period in Cashier. If you don't want that you need to add that behavior by listening to the incoming webhook with https://laravel.com/docs/8.x/billing#defining-webhook-event-handlers

PR #1316, opened 2022-02-05
Fix subscription validity if cancelled during free trial period

kpolicar:
fixes https://github.com/laravel/cashier-stripe/issues/1315


taylorotwell:
I'm a little concerned this could break existing applications that let users keep using the application until the end of their trial period in this scenario? /cc @driesvints

driesvints:
As explained in #1315 this expected because there's a grace period to take into account.

PR #1324, opened 2022-02-20
docs: rawDiscountFor returns first() StripeObject from collection

markkimsal:
`rawDiscountFor` returns a StripeObject or null, not an int.  The usage of the return value as an object is displayed in `discountFor`

```php
        if (! is_null($discountAmount = $this->rawDiscountFor($discount))) {
            return $this->formatAmount($discountAmount->amount);
        }
```


driesvints:
Heya, I sent in an alternative PR that'll fix the method's return value instead. Thanks!

https://github.com/laravel/cashier-stripe/pull/1325

Issue #1326, opened 2022-02-22
"Stripe\Exception\InvalidRequestException: Invalid integer: in file ..." swapping a subscription item with metered price


pietrantonio91:
- Cashier Version: 13.7.0
- Laravel Version: 8.73.2
- PHP Version: 7.4.27
- Database Driver & Version: MySQL 8

### Description:
When swapping a suscription item price of an existing subscription it retrieve this notice because of quantity set to null:
> Stripe\Exception\InvalidRequestException: Invalid integer: in file ...

I figured out these should be the problem lines:
[https://github.com/laravel/cashier-stripe/blob/6422206a8cdc986f227f2e4a140073801759adcc/src/SubscriptionItem.php#L147-L153](https://github.com/laravel/cashier-stripe/blob/6422206a8cdc986f227f2e4a140073801759adcc/src/SubscriptionItem.php#L147-L153)

I solved this issue like this:
```
$stripeSubscriptionItem = $this->updateStripeSubscriptionItem(array_merge(
            array_filter([
                'price' => $price,
                'quantity' => $this->quantity,
                'payment_behavior' => $this->paymentBehavior(),
                'proration_behavior' => $this->prorateBehavior(),
                'tax_rates' => $this->subscription->getPriceTaxRatesForPayload($price),
            ], function($value) {
                return $value !== null;
            }), 
        $options));
```

### Steps To Reproduce:
1. Create a subscription with a metered price
2. `$subscription_item->swap('price_NewPriceId');`

driesvints:
Thanks for reporting and the PR!

PR #1328, opened 2022-02-22
[13.x] Fix swapping metered price of subscription item to existing su…

pietrantonio91:
Fix swapping metered price (with quantity null) to existing subscription. Stripe don't accept null value for quantity (integer required) so it has to be removed from request if it's null.

Fix for #1326


driesvints (approved):

driesvints:
Thanks!

PR #1349, opened 2022-04-22
Add Stripe Checkout locale option.

markoman78:
Add Stripe Checkout locale option to Cashier config / env file.


driesvints:
You can already pass this with the options array. Is there any reason why you want this as a config option?

markoman78:
> You can already pass this with the options array. Is there any reason why you want this as a config option?

Didn't find any information about of changing Stripe Checkout language other than it's "auto" mode (Browser's language). Could this be more clear when it can be changed from config?

driesvints:
I don't think we should change this. By default it's set to `auto` so the user's browser language will be used. This PR forces it to `english`. I also think it's not worth to add this as an option in Cashier as you can already simply pass it with the options array.

https://stripe.com/docs/api/checkout/sessions/create#create_checkout_session-locale

PR #1363, opened 2022-05-12
[13.x] Fix reference to billable

driesvints:
This should reference the billable rather than the Cashier class.


sts-ryan-holton:
@driesvints This is an interesting change, noticing it's related to Cashier webhook? What were the potential issues with this prior to the change (if there were any)?

driesvints:
It's always better to reference a billable object since you may overwrite the `stripe` method on your model for some reason (although I don't recommend doing that).

PR #1367, opened 2022-05-21
Allow the ability to set different API versions of the Stripe API

timleland:
I am wanting to be able to change the stripe API version to add a beta option and currently, I do not think there is a way to customize STRIPE_VERSION. Please let me know If there is a way to set the stripe API version and I will close this pr.

Thanks!

### DIFF
--- a/config/cashier.php
+++ b/config/cashier.php
@@ -19,6 +19,8 @@ return [
 
     'secret' => env('STRIPE_SECRET'),
 
+    'stripe_version' => env('STRIPE_VERSION'),
+
     /*
     |--------------------------------------------------------------------------
     | Cashier Path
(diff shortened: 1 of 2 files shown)


driesvints:
Cashier is only compatible with a specific stripe api version, this to prevent breakages with the current implementation of the stripe api. We upgrade this version in New major releases. 

Why do you need this if I may ask?

driesvints:
FYI: you can always override this by overwriting the stripe method in your billable model

timleland:
Stripe is doing a beta and to enable it you have to add a **2020-08-27;link_beta=v1** to the end of the Stripe API version. Could you share an example of how to override the stripe method?

Thanks!

driesvints:
```
    /**
     * Get the Stripe SDK client.
     *
     * @param  array  $options
     * @return \Stripe\StripeClient
     */
    public static function stripe(array $options = [])
    {
        return Cashier::stripe(array_merge(['stripe_version' => '2020-08-27;link_beta=v1'], $options));
    }
```

timleland:
Thanks!

PR #1369, opened 2022-05-25
[13.x] Extract Stripe ID features to `BelongsToStripe` trait, allowing `stripe_id` key customization

stevebauman:
## Description

This PR extracts existing Stripe ID methods to another trait, so developers can apply this same trait to other models that they may want to synchronize in Stripe, such as Products.

It will provide a shortcut for developers to leverage a Cashier-like API with their own implementations.

This PR also provides a way for developers to alter the `stripe_id` column in the `users` update table migration, as well as in the `subscriptions` and `subscription_items` table migrations. Once updated in the migration, the users may set the `$stripeKey` property on the model to reference its new attribute location.

## Examples

**Custom Product Model**:

```php
namespace App\Models;

use Laravel\Cashier\Concerns\BelongsToStripe;

class Product extends Model
{
    use BelongsToStripe;

    public function createAsStripeProduct()
    {
        $product = $this->stripe()->products->create(['name' => $this->name]);

        $this->setStripeId($product->id);

        $this->save();

        return $product;
    }
}
```

**Custom Stripe Key**:

```php
class User extends Model
{
    use Billable;

    protected $stripeKey = 'stripe_customer_id';
}
```

Let me know your thoughts! No hard feelings on closure ❤️

### DIFF
--- a/database/factories/SubscriptionFactory.php
+++ b/database/factories/SubscriptionFactory.php
@@ -31,7 +31,7 @@ class SubscriptionFactory extends Factory
         return [
             (new $model)->getForeignKey() => ($model)::factory(),
             'name' => 'default',
-            'stripe_id' => 'sub_'.Str::random(40),
+            (new $model)->stripeKey() => 'sub_'.Str::random(40),
             'stripe_status' => StripeSubscription::STATUS_ACTIVE,
             'stripe_price' => null,
             'quantity' => null,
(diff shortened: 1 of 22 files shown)


driesvints:
Hi @stevebauman. I'm going to respectfully decline this as I think this vastly complicates the codebase for us to maintain.

If we go this road I'd rather adopt the same approach we use in Cashier Paddle so multiple models can be a billable. But unfortunately that approach is a major breaking change which we've been postponing. I do hope that at some point we could roll it out but for now we're going to keep Cashier to a single billable model.

stevebauman:
Ok no worries! 👍 Thanks for the reply

PR #1380, opened 2022-06-17
Add new method trialMonths

padre:
There are a lot of times that our customers have several months of trial period. I always have to use trialUntil():

### Before:

```
$user->newSubscription('default', 'price_monthly')
            ->trialUntil(Carbon::now()->addMonths(3))
            ->create($paymentMethod);
```
### After

```
$user->newSubscription('default', 'price_monthly')
           ->trialMonths(3)
            ->create($paymentMethod);
```

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -161,6 +161,19 @@ class SubscriptionBuilder
 
         return $this;
     }
+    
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think `trialUntil` pretty much solves this situation.

PR #1381, opened 2022-06-20
feat: allow to get stripe customer with additional parameters

1davidnaj:
allow to get stripe customer with additional parameters


driesvints:
This is unfortunately a breaking change to the method signature. Please send this to the master branch.

driesvints:
Which attributes did you want to expand?

1davidnaj:
For example 'sources' attribute, because without it there is no access to customer sources

PR #1389, opened 2022-07-04
add useStripeClient method to make StripeClient mockable

ahmedash95:
Hello, 

I noticed in the documentation it says [Cashier hits stripe api](https://laravel.com/docs/9.x/billing#testing:~:text=When%20testing%20an%20application%20that%20uses%20Cashier%2C%20you%20may%20mock%20the%20actual%20HTTP%20requests%20to%20the%20Stripe%20API) with testing secrets and it's recommended way.  and If I want to mock I should [partially re-implement Cashier's own behavior](https://laravel.com/docs/9.x/billing#testing:~:text=partially%20re%2Dimplement%20Cashier%27s%20own%20behavior). 

so for my case, I had to implement a fake user class to do that like below for every case.
```php
class UserWithStripeFake extends User
{
    protected $table = 'users';

    protected $primaryKey = 'id';

    public static function stripe(array $options = [])
    {
        return Mockery::mock(StripeClient::class)->expects('request')->andReturn((object)[
            'status' => 'past_due',
        ]);
    }
}
```
but It would be much nicer if Cashier provides a way to make it easy to mock/replace `StripeClient` class. so in this PR I'm allowing this by following an existing convention to override [User model](https://github.com/laravel/cashier-stripe/blob/13.x/src/Cashier.php#L216) and [Subscription model](https://github.com/laravel/cashier-stripe/blob/13.x/src/Cashier.php#L227), that makes it easy to just have the code like below
```php
$stripeClient = $this->mock(StripeClient::class)->expects('request')->andReturn((object)[
    'status' => 'past_due',
]);

Cashier::useStripeClient($stripeClient);
```

### DIFF
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -255,6 +255,6 @@ class Cashier
      */
     public static function useStripeClient($client)
     {
-        self::$stripeClientInstance = $client;
+        static::$stripeClientInstance = $client;
     }
 }
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -91,6 +91,13 @@ class Cashier
     public static $subscriptionItemModel = SubscriptionItem::class;
 
     /**
+     * The stripe client instance.
+     *
+     * @var StripeClient
+     */
+    public static $stripeClientInstance = null;
+
+    /**
      * Get the customer instance by its Stripe ID.
      *
      * @param  \Stripe\Customer|string|null  $stripeId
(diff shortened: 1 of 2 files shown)


driesvints:
Overwriting the `stripe` method is the recommended way. I don't think that adding a static property here makes sense tbh. In general, I suggest you to look into stripe-mock: https://github.com/stripe/stripe-mock

PR #1391, opened 2022-07-05
[12.x] Allow dompdf v2 and money v4

GrahamCampbell:


driesvints:
What are the breaking changes here?

GrahamCampbell:
I don't think there are any that affect us. https://github.com/dompdf/dompdf/wiki/Migration-Guide

PR #1393, opened 2022-07-11
[13.x] Allow dompdf v2

carlalexander:
This is to add the change from #1391 to the `13.x` branch. The last version of domdpf v1 has a [security vulnerability](https://github.com/advisories/GHSA-pf6p-25r2-fx45) and it doesn't look like they intend to patch v1.


mantebalzaraviciute:
@carlalexander

PR #1405, opened 2022-07-28
[14.x] Add method to reset default payment method properties

mpskovvang:
The PR allows developers to customize behaviour when default payment method properties are being reset.

In our case, we do not use the `pm_type` and `pm_last_four` model properties. Instead we have a `stripe_default_payment_method` property.

We can already overwrite how payment method details are filled with:
```php
protected function fillPaymentMethodDetails($paymentMethod)
{
    $this->forceFill([
        'stripe_default_payment_method' => $paymentMethod->id ?? null,
    ]);

    return $this;
}
```

This change also allows updating the model when the default payment method is being removed:
```php
protected function resetDefaultPaymentMethod()
{
    $this->forceFill([
        'stripe_default_payment_method' => null,
    ])->save();
        
    return $this;
}
```

I'm targeting the next major version because this could be a breaking change if the method is already present in the billable model.

### DIFF
--- a/src/Concerns/ManagesPaymentMethods.php
+++ b/src/Concerns/ManagesPaymentMethods.php
@@ -113,10 +113,7 @@ trait ManagesPaymentMethods
 
         // If the payment method was the default payment method, we'll remove it manually...
         if ($stripePaymentMethod->id === $defaultPaymentMethod) {
-            $this->forceFill([
-                'pm_type' => null,
-                'pm_last_four' => null,
-            ])->save();
+            $this->resetDefaultPaymentMethod();
         }
     }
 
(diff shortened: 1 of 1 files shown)


driesvints:
I'm not sure why you'd need this method to reset these values if you're not using them anyway? Just keep on using your own property?

PS. this isn't a breaking change imo.

mpskovvang:
The reset method is helpful to avoid Cashier trying to fill missing database columns.

If a user removes the default payment method by calling `$user->deletePaymentMethod(...)`, an SQL error will occur:
`Undefined column: 7 ERROR:  column "pm_type" of relation "users" does not exist`.

The PR allows overriding this behaviour.

My current workaround is to override the `setAttribute` method on the billable model to ignore the `pm_type` and `pm_last_four` attributes.

PS. do you happen to know a good article about breaking changes definition? 🙂

driesvints:
> Undefined column: 7 ERROR: column "pm_type" of relation "users" does not exist.

This error cannot happen with Cashier Stripe as it's required to have this column on your customers table. We do not support removing it.

> PS. do you happen to know a good article about breaking changes definition? 🙂

Not really sorry... the reason why this isn't a breaking change is because there's no change in behavior. You're simply extracting it to a different method.

PR #1411, opened 2022-08-03
Allow stripe php sdk v9

ankurk91:
https://github.com/stripe/stripe-php/blob/master/CHANGELOG.md


driesvints:
Thanks. I've resubmitted this myself so the full test suite can run. At first glance on the changelog, there don't seem to be breaking changes to Cashier v13 or Spark v2.

PR #1416, opened 2022-08-11
[14.x] Drop MoneyPHP 3.x

driesvints:
Drops MoneyPHP 3.x support for Cashier v14


taylorotwell:
@driesvints do we have to drop 8.0 support?

driesvints:
@taylorotwell not necessarily if you don't want to. I'll reverse.

PR #1420, opened 2022-08-12
[14.x] New subscription behavior

driesvints:
This PR adopts the new recommended `default_incomplete` payment behavior that Stripe now recommends. The main adjustments here are the transition to the new `default_incomplete` value when passing the `payment_behavior` when creating or updating subscriptions or subscription items and confirming the payment intents server-side before throwing the `IncompletePayment` exception.

I've extracted the logic to handle payment failures in a new `HandlesPaymentFailures` trait. In here, we'll check if a subscription has a failed payment. If so, we'll check if it needs confirmation and when it does we'll confirm it server-side. If a payment method has been provided as well (usually when creating a subscription) we'll pass it along. After confirming the payment we'll need to update the subscription status. If no confirmation is needed and there is an incomplete payment we'll cascade the exception immediately. 

I managed to made this PR backwards compatible for existing integrations. Existing integrations are expected to handle failing payments anyway so it should be a smooth upgrade.

The next step I want to take is to update our docs with the flow of how Stripe recommends to handle collecting payment details which is after a subscription has been already created.

More info: https://stripe.com/docs/billing/subscriptions/build-subscriptions?ui=elements#create-subscription

### DIFF
--- a/src/Concerns/InteractsWithPaymentBehavior.php
+++ b/src/Concerns/InteractsWithPaymentBehavior.php
@@ -11,7 +11,19 @@ trait InteractsWithPaymentBehavior
      *
      * @var string
      */
-    protected $paymentBehavior = StripeSubscription::PAYMENT_BEHAVIOR_ALLOW_INCOMPLETE;
+    protected $paymentBehavior = StripeSubscription::PAYMENT_BEHAVIOR_DEFAULT_INCOMPLETE;
+
+    /**
+     * Set any new subscription as incomplete when created.
+     *
+     * @return $this
+     */
+    public function defaultIncomplete()
+    {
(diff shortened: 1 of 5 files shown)


src/Concerns/InteractsWithPaymentBehavior.php
```diff
@@ -11,7 +11,19 @@ trait InteractsWithPaymentBehavior
      *
      * @var string
      */
-    protected $paymentBehavior = StripeSubscription::PAYMENT_BEHAVIOR_ALLOW_INCOMPLETE;
+    protected $paymentBehavior = StripeSubscription::PAYMENT_BEHAVIOR_DEFAULT_INCOMPLETE;
```
driesvints, line 14: This is the main change of this PR.

PR #1435, opened 2022-09-13
[14.x] Configurable Webhook Enabled Events

yob-yob:
This will allow adding more enabled webhook events when creating a stripe webhook using `php artisan cashier:webhook`


driesvints (approved):
I don't see a big problem with this. Can you also send in a PR to the docs to document this?

yob-yob:
I'm sorry for the late reply, I will try to construct a PR to documentation for this. Thank you thou.

PR #1442, opened 2022-09-26
On subscription cancel trial_ends_at keeps

bkd27:
The case when the user subscribed with a trial period and the subscription is canceled then the 'trial_ends_at' value stays there, which causes the check subscription status $user->subscribed('default'))  return true even if the subscription is permanently canceled.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -1073,7 +1073,7 @@ class Subscription extends Model
         $this->fill([
             'stripe_status' => StripeSubscription::STATUS_CANCELED,
             'ends_at' => Carbon::now(),
-            'trial_ends_at' => NULL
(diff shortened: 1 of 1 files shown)


driesvints:
This is expected. It's to not end the generic trial they might have. If you want this to explicitly be ended use:

```
$subscription->skipTrial()->cancelNow();
```

bkd27:
Hi Dries,

Yes, that can be only if the subscription is cancelled via code, What if the subscription is cancelled via Stripe dashboard or Stripe customer portal, As in webhook there not that case.

driesvints:
Yeah that's right. If you cancel through those, the generic trial won't be skipped. I'll think on this a bit and might send something in to make this configurable.

PR #1444, opened 2022-09-26
[14.x] Skip trials when cancelling subscriptions from dashboard

### LINKED ISSUE #1443: Disable subscription entirely when canceling through Stripe Dashboard
When a subscription is cancelled through the Stripe Dashboard or Customer Portal, it's not immediately ended when the subscription is still on a generic trial. See https://github.com/laravel/cashier-stripe/pull/1442.

Maybe a config option for this could be added or a simpler way of setting this.

bkd27:

### DIFF
--- a/config/cashier.php
+++ b/config/cashier.php
@@ -55,6 +55,7 @@ return [
             'invoice.payment_action_required',
             'invoice.payment_succeeded',
         ],
+        'skip_trial' => env('STRIPE_WEBHOOK_SKIP_TRIAL', false),
     ],
 
     /*
--- a/config/cashier.php
+++ b/config/cashier.php
@@ -55,7 +55,6 @@ return [
             'invoice.payment_action_required',
             'invoice.payment_succeeded',
         ],
-        'skip_trial' => env('STRIPE_WEBHOOK_SKIP_TRIAL', false),
     ],
 
     /*
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -70,6 +70,13 @@ class Cashier
     public static $calculatesTaxes = false;
 
     /**
+     * Indicate if Cashier should skip (generic) trials when cancelling subscriptions.
+     *
+     * @var bool
+     */
+    public static $skipTrials = false;
+
+    /**
      * The default customer model class name.
      *
      * @var string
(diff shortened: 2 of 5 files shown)


driesvints:
I'll have a look later at this.

driesvints:
I did some changes. Now when you add `Cashier::skipTrials()` to your service provider, Cashier Stripe will always skip the (generic) trial when cancelling subscriptions.

driesvints (approved):

driesvints:
@bkd27 while working on this I realized we're modifying subscription trials here but what you actually want is to end a customer's generic trial on their user record. Generic trials are a Cashier level thing and cannot be controlled from Stripe. You'll have to implement some custom code using webhooks somehow to get the behavior that you desire.

As for subscription trials you can already end them whenever you wish when cancelling them from the dashboard:

Screenshot 2022-09-26 at 16 05 44

bkd27:
> @bkd27 while working on this I realized we're modifying subscription trials here but what you actually want is to end a customer's generic trial on their user record. Generic trials are a Cashier level thing and cannot be controlled from Stripe. You'll have to implement some custom code using webhooks somehow to get the behaviour that you desire.
> 
> As for subscription trials you can already end them whenever you wish when cancelling them from the dashboard:
> 
> Screenshot 2022-09-26 at 16 05 44

Hi, dries

No, the generic trial is a purely separate thing (That is customer level trial), I am concerned about the subscription trial only, If a subscription is cancelled 'Immediately' in a stripe or via the customer portal, the default behaviour is to if the subscription has a trial on then that should be removed. If that is not then what is the use of cancelling the subscription immediately, 

The code that I added controlling that behaviour via config will be handy for that. 

Let me know if you have further questions on this.

driesvints:
I'm still investigating this. Contacted Stripe and will reply once I know more.

driesvints:
I didn't get to this before my vacation but will pick it up again afterwards.

driesvints:
Hi there, I finally found time to look into this again. I've decided to send in a PR that changes the current behavior in the next major version: https://github.com/laravel/cashier-stripe/pull/1465

While this doesn't help you right now, in the next version it will work as you'd expect. I think also that we shouldn't add an option like this PR to accommodate for a bridge in between the versions so unfortunately you'll need to wait a bit longer before it's changed.

PR #1448, opened 2022-10-05
Fix undefined property id notice

Lyrisbee:
Fix `PHP message: Stripe Notice: Undefined property of Stripe\Invoice instance: id` when fetching upcoming invoice data


src/Invoice.php
```diff
@@ -569,7 +569,7 @@ protected function refreshWithExpandedData()
             'total_tax_amounts.tax_rate',
         ];
 
-        if ($this->invoice->id) {
+        if (property_exists($this->invoice, 'id')) {
```
driesvints, line 572: We'll need to check for a value as well.

PR #1449, opened 2022-10-06
Fix StripeObject values problem

Lyrisbee:
StripeObject saves the value to the array, not a property. This PR also fixes tests.

https://github.com/stripe/stripe-php/blob/e73d05039943e3bbcd0f068f220ee55f88e8952c/lib/StripeObject.php#L152-L160

test result:
圖片


driesvints:
Thank you!

PR #1459, opened 2022-10-25
Add invoice pay options as param.

altin7m:
Add invoice pay options as param when creating an invoice for a customer. Pay options are optional.


driesvints:
We can't add new arguments to method signatures because that's breaking. I'll have a look at this as soon as I find the time.

driesvints:
Heya, I sent in an alternative non-breaking PR to https://github.com/laravel/cashier-stripe/pull/1464.

Issue #1454, opened 2022-10-11
Option to keep incomplete subscriptions active


anthonycreek:
I've run into an issue where a user could have an existing subscription and then change the credit card to one that requires SCA. When the subscription renews and Stripe billing charges the card, the subscription status will be updated to `incomplete` since they need to complete the SCA. The same could happen when swapping subscriptions as well. 

At that point, the model is no longer subscribed according to Cashier since the [active method](https://github.com/laravel/cashier-stripe/blob/1a86a6ccc141c8321763eb02cbfb24fc58ca3fa0/src/Subscription.php#L227) returns `false`. 

In our case, that means the entire team immediately loses access to the workspace. Ideally we'd have the option to treat this like we do a `past_due` subscription using `Cashier::keepPastDueSubscriptionsActive()`. 

So my suggestion is the addition of `Cashier::keepIncompleteSubscriptionsActive()` so that we can give the user time to complete SCA - especially if it happens when they aren't actively using the app (upon renewal with a new card). If they don't complete SCA then the subscription will cancel in a few days based on the Stripe settings. 

I can create a PR for this if it is of interest and there aren't any gotchas I should be aware of here.

driesvints:
I thought about this one quite a bit but I understand your predicament. Even with the thought that I'd never go down this path myself I can see others doing it. Just be mindful with this because activating it means your customers are basically using your product for free.

See https://github.com/laravel/cashier-stripe/pull/1466

anthonycreek:
Nice, thank you! We have Stripe rules set up to cancel the subscription after 7 days if they don't get the payment made so we'll grab the webhook and cut off access at that point. Appreciate the update here though. Cleans up our code a bit where we were working around it.

PR #1466, opened 2022-10-28
[14.x] Allow incomplete subscriptions as active

driesvints:
This PR allows subscriptions to stay active as long as they're in an `incomplete` status. The reason for this is that it could be that subscriptions are pending on SCA confirmation etc. During this window, customers could miss out on any access to the product they're using. In normal situations, that's what you want since at this point, they're not paying for the product. However, often this can be disruptive as well to the customer experience. Therefor allowing apps to keep these incomplete subscriptions as active is a valid option to provide. Works exactly the same as `keepPastDueSubscriptionsActive`.

The default behavior is unchanged as it should be. As soon as a subscription hits `incomplete_expired` the subscription is no longer active.

https://github.com/laravel/cashier-stripe/issues/1454

### DIFF
--- a/src/Cashier.php
+++ b/src/Cashier.php
@@ -63,6 +63,13 @@ class Cashier
     public static $deactivatePastDue = true;
 
     /**
+     * Indicates if Cashier will mark incomplete subscriptions as inactive.
+     *
+     * @var bool
+     */
+    public static $deactivateIncomplete = true;
+
+    /**
      * Indicates if Cashier will automatically calculate taxes using Stripe Tax.
      *
      * @var bool
@@ -191,6 +198,18 @@ class Cashier
     public static function keepPastDueSubscriptionsActive()
     {
         static::$deactivatePastDue = false;
+
+        return new static;
+    }
+
+    /**
+     * Configure Cashier to maintain incomplete subscriptions as active.
+     *
+     * @return static
+     */
+    public static function keepIncompleteSubscriptionsActive()
+    {
+        static::$deactivateIncomplete = false;
 
         return new static;
     }
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -217,7 +217,7 @@ class Subscription extends Model
     public function active()
     {
         return ! $this->ended() &&
-            $this->stripe_status !== StripeSubscription::STATUS_INCOMPLETE &&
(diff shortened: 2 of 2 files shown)


magarrent:
Hey, @driesvints! Thanks for the PR

I think there is another problem with the active/incomplete subscriptions.

(I'm using Stripe checkout + Webhooks) - Cashier v14.3.0

I've just noticed that the _customer.subscription.created_ event, always returns an _"inactive"_ status. Then you have to process the _customer.subscription.updated_ to notice that the payment is complete and the subscription is active.
But I think that the webhooks go as fast as the server can't get the current subscription when they receive the _customer.subscription.updated_ event.
The subscription remains **incomplete** in the _subscriptions table_ always, including after _customer.subscription.updated_ event.
**But** if you run again the _customer.subscription.updated_ event, it will successfully be set as "active", which means that the webhook is working well.


I saw your PR and I thought that it will solve this but It does not in my case, should I active this option somewhere?

Thank you, amazing job here!

driesvints:
> I've just noticed that the customer.subscription.created event, always returns an "inactive" status. 

You mean "incomplete"?

> But I think that the webhooks go as fast as the server can't get the current subscription when they receive the customer.subscription.updated event.

Webhooks are asynchronous so that means the updated event can arrive before the created event. It's unfortunate but we really can't do anything about that :(

> I saw your PR and I thought that it will solve this but It does not in my case, should I active this option somewhere?

This PR has nothing to do with your problem and won't help.

I totally get the pickle you're in but right now I can't think of anything we can do without building an entire state machine and delay webhook processing. Try contacting Stripe to see what they have to say?

magarrent:
Yeah sorry, incomplete! I reached the Stripe support and this is the response from the agent:

>  Thanks for waiting, Marc. After some investigation, I've found that you are creating subscriptions with Checkout Client Only. 


> Have you any examples of the subscription-created event showing a different status than incomplete? That actually would be expected behaviour.
> The subscription is generated as incomplete, then generates an invoice as open, which generates a payment intent as incomplete. When the payment intent is succeeded, the invoice transitions to paid, and the subscription to active.

So if the webhooks are completely asynchronous how can we update the subscription to active? :S

I'm using the Test environment but hope this won't take any effect.

driesvints:
Hi @magarrent. Unfortunately I have no answer to this. It's best that you contact Stripe again and urge them to give you a proper answer to how this should be dealt with.

PR #1473, opened 2022-11-11
Use empty() instead of is_null() to check for empty stripe ID

vantezzen:
Cashier currently regards `""` (empty string) and `false` as valid Stripe IDs. This results in methods like `$user->createOrGetStripeCustomer();` returning a Stripe exception for `"The resource ID cannot be null or whitespace"` instead of creating a stripe user.
Cashier itself uses `NULL` to indicate an empty value, depending on how the data is stored or how other scripts process it, the value could possible another empty-type value. This makes it hard to debug the reason Cashier is not creating a new stripe user.

To achieve this, this PR changes the `is_null` check to `empty` to also regard other empty-like types as an empty Stripe ID - including `null`, making a separate `is_null` check unneeded.

### DIFF
--- a/src/Concerns/ManagesCustomer.php
+++ b/src/Concerns/ManagesCustomer.php
@@ -32,7 +32,7 @@ trait ManagesCustomer
      */
     public function hasStripeId()
     {
-        return ! is_null($this->stripe_id);
+        return ! empty($this->stripe_id);
     }
 
     /**


driesvints:
This value needs to be either a filled out string or `null` is there isn't any. If there's an outside script filling out these values then it contains a bug and needs to be updated for the expected value.

PR #1478, opened 2022-11-28
[14.x] Update Stripe API version and Stripe PHP SDK v10 support

driesvints:
Usually updating the Stripe API version is done in major releases of Cashier. But since the changes are minimal here I think it's safe to do them in a minor release this time.


ankurk91:
Please allow stripe php sdk version 10 as well

driesvints:
Thanks @ankurk91. Added that.

PR #1479, opened 2022-12-02
[14.x] Cursor pagination for invoices

driesvints:
This PR adds cursor pagination for customer invoices. Using the `starting_after` and `ending_before` fields on [the invoices list API endpoint](https://stripe.com/docs/api/invoices/list) by Stripe, we can easily paginate through a customer's invoices using cursors.

```php
$paginator = $billable->cursorPaginateInvoices();
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "illuminate/http": "^9.21",
         "illuminate/log": "^9.21",
         "illuminate/notifications": "^9.21",
+        "illuminate/pagination": "^9.21",
         "illuminate/routing": "^9.21",
         "illuminate/support": "^9.21",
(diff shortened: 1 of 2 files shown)


mbardelmeijer:
Noticed that this method includes pending invoices by default (see https://github.com/laravel/cashier-stripe/pull/1479/files#diff-720929e0dcf6c94b9991fa4a61163a9cb31ac38960a58e2b5d71dc0e3571ef3fR332). Is there a way to disable this to only get open & paid invoices for instance?

driesvints:
Yeah it's basically "everything" which is intended. Unfortunately the Stripe API has no way to list multiple types of invoices. Just one specific type or every type can be retrieved at once.

Issue #1488, opened 2023-01-09
Create subscription with both metered and regular prices


talelmishali:
- Cashier Version: 14.6.0
- Laravel Version: 9.46.0
- PHP Version: 8.2.0
- Database Driver & Version: MySQL

### Description:
When trying to create a new subscription with both metered and regular prices, Stripe will throw ``invalid_request_error - quantity``. It appears that when creating a new subscription and passing a price with no quantity, the builder will automatically add ``1`` as the default quantity, which results in an error. 

When calling ``meteredPrice()``, the quantity parameter is passing ``null``, which pushes the price correctly into ``items``. 

### Steps To Reproduce:
1. Setting up metered and regular prices on Stripe.
2. Creating a customer
3. Creating a new subscription with both metered and regular prices.

```php
$this->user
            ->newSubscription('default', $prices)
            ->trialDays($trialDays)
            ->create($paymentMethod);
```

#### What I push as prices
```php
array:2 [ 
  0 => "price_1MO063L7Q3I087mfWiVUvPnp",
  1 => array:2 [
    "price" => "price_1MO24mL7Q3I087mfNsfFjJ4H",
    "quantity" => 1
  ]
]
```
#### ⚠️ What the builder push to Stripe
```php
  #items: array:2 [
    "price_1MO063L7Q3I087mfWiVUvPnp" => array:2 [
      "price" => "price_1MO063L7Q3I087mfWiVUvPnp",
      "quantity" => 1
    ],
    "price_1MO24mL7Q3I087mfNsfFjJ4H" => array:2 [
      "price" => "price_1MO24mL7Q3I087mfNsfFjJ4H",
      "quantity" => 1
    ]
  ]
```

#### ✅ What the builder push when using ``meteredPrice()``
```php
  #items: array:2 [
    "price_1MO24mL7Q3I087mfNsfFjJ4H" => array:2 [
      "price" => "price_1MO24mL7Q3I087mfNsfFjJ4H",
      "quantity" => 1
    ],
    "price_1MO063L7Q3I087mfWiVUvPnp" => array:1 [
      "price" => "price_1MO063L7Q3I087mfWiVUvPnp"    -----------------> No quantatity on meterd price.
    ]
  ]
```

#### ⚠️ Stripe log for an incorrect request 
```bash
invalid_request_error - quantity
You cannot set the quantity for metered plans.

  "items": {
    "0": {
      "quantity": "1",
[...]

PR #1489, opened 2023-01-10
[14.x] Allow metered prices in subscription builder

driesvints:
This PR allows metered prices to be added to the subscription builder's construct. This means users are no longer obligated to explicitly use the `meteredPrice` on the subscription builder. 

This PR technically contains breaking changes to method signatures. However, I do not believe people are overwriting these. If you want, I can send this to master instead.

Closes https://github.com/laravel/cashier-stripe/issues/1488

### DIFF
--- a/tests/Unit/SubscriptionBuilderTest.php
+++ b/tests/Unit/SubscriptionBuilderTest.php
@@ -18,8 +18,8 @@ class SubscriptionBuilderTest extends TestCase
         ]);
 
         $this->assertSame([
-            'price_foo' => ['price' => 'price_foo', 'quantity' => 1],
-            'price_bux' => ['price' => 'price_bux', 'quantity' => 1],
+            'price_foo' => ['price' => 'price_foo'],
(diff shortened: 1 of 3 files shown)


tests/Unit/SubscriptionBuilderTest.php
```diff
@@ -18,8 +18,8 @@ public function test_it_can_be_instantiated_with_both_normal_and_metered_prices(
         ]);
 
         $this->assertSame([
-            'price_foo' => ['price' => 'price_foo', 'quantity' => 1],
-            'price_bux' => ['price' => 'price_bux', 'quantity' => 1],
+            'price_foo' => ['price' => 'price_foo'],
+            'price_bux' => ['price' => 'price_bux'],
```
driesvints, line 22: The change to this test demonstrates that the expected behavior.

driesvints:
Realized it's not possible to know which price is metered and which one is recurring. Recurring prices need a quantity. So unfortunately this is a no-go.

Issue #1491, opened 2023-01-10
no findSetupIntent method


quantumwebco:
With the newer Stripe payment flows with Elements and Payment Request Buttons, ```stripe.confirmSetup()``` redirects you to a specified url with the setup intent in the query string. To get the payment method you have to retrieve the setup intent and get the payment method ID from the setup intent properties

We can create a setup intent with ```$user->createSetupIntent()``` but unable to retrieve it except using the SDK directly

Can I add a method ```findSetupIntent($setupIntentId)``` and maybe even ```getPaymentMethodFromSetupIntent($setupIntentId)``` to the ```ManagesPaymentMethods``` concern for convenience?

driesvints:
Heya, can you link to this guide so I can check it out?

quantumwebco:
Here's the one for setupIntents https://stripe.com/docs/payments/save-and-reuse 

You'll notice it uses a _payment_ element and ```confirmSetup()``` rather than a simple _card_ element and ```confirmCardSetup()```. ```confirmSetup()``` takes an object of options, the setupIntent client secret and a redirect URL, this is so it can handle all the SCA stuff and payment methods that require a redirect. I have only enabled _cards_ (card/gpay/applepay) as I'm unsure what would happen setting something like Klarna as a default payment method

The Payment Request Button is quite different https://stripe.com/docs/stripe-js/elements/payment-request-button?locale=en-GB
I'm haven't attempted an implementation yet but it does use ```confirmCardPayment()``` like a card element can, but I'm not sure if it works with a setupIntent and and saving the payment method for future use

driesvints:
I think we can add a `findSetupIntent` with an `$option` argument that allows you to expand the payment method if you like. Feel free to send in a PR 👍

PR #1505, opened 2023-02-22
[14.x] Provide findSetupIntent method

andycowan:
The new Payment element requires a redirect to a new page after submitting the payment details, to enable the insertion of intermediate paymentMethod steps (for example bank pages etc.)

When the return_url is reaches, the setupIntent is supplied as a parameter, with the purpose of allowing the code to retrieve the setupIntent, to get the payment_method id.

This PR provides a findSetupIntent method on the billable entity, to enable retrieval of the setupIntent.

Test included.

This is described by @quantumwebco in issue https://github.com/laravel/cashier-stripe/issues/1491

fixes https://github.com/laravel/cashier-stripe/issues/1491

### DIFF
--- a/src/Concerns/ManagesPaymentMethods.php
+++ b/src/Concerns/ManagesPaymentMethods.php
@@ -23,6 +23,19 @@ trait ManagesPaymentMethods
     }
 
     /**
+     * Retrieve a SetupIntent from Stripe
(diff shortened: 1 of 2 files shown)


driesvints:
Thanks!

PR #1523, opened 2023-04-06
add addOrIncrementPrice method to Subscription

endel:
Hi guys, does this addition make sense? What are your thoughts?

I expected to just use a method like that but had to come up with this simple implementation.


taylorotwell:
Hey there - I suggest adding this to your own application.

PR #1525, opened 2023-04-11
[14.x] Add ability to set a description on "one-off" charge checkouts

AmrAb-05:
Right now its not possible to set a description for checkout sessions created with for example
```
$request->user()->checkoutCharge(1200, 'T-Shirt', 5);
```


AmrAb-05 (commented):
How does this look?

driesvints (approved):

Issue #1526, opened 2023-04-11
Delete Stripe Customer functionality missing


rkeshwani:
This library is missing the ability to delete a stripe customer.
So for example, if a customer requests for their account to be deleted, you currently can't delete their information in Stripe via Laravel Cashier.
stripe-php's Customer class contains a DELETE trait that exposes a delete() method which should be able to perform this operation.

driesvints:
Thanks for the PR

PR #1527, opened 2023-04-11
Added Stripe Delete Customer method on ManageCustomer

rkeshwani:
This adds a method to the ManagesCustomer Concern called deleteStripeCustomer which deletes the customer from stripe and nulls the reference to that user in the current object.
This is helpful for when you need to delete the stripe customer.

I've provided tests but am unable to run them as my account is not configured to run all of the tests.
For implementation:
It should be calling the following method:
https://github.com/stripe/stripe-php/blob/b7744ede8a959b06f05859da10bdfbe561411267/lib/Service/CustomerService.php#L209

Based on this documentation, there is no need for parameters beyond the customer id.
https://stripe.com/docs/api/customers/delete

It shouldn't break anything because this is a new feature and no current methods have been changed. But I have not tested to see if it works because I hesitate to put in my stripe id and run ALL of the tests.

Issue Reference: #1526

### DIFF
--- a/src/Concerns/ManagesCustomer.php
+++ b/src/Concerns/ManagesCustomer.php
@@ -124,6 +124,21 @@ trait ManagesCustomer
     }
 
     /**
+     * Delete the Stripe customer for the current user.
+     * @return void
+     */
+    public function deleteStripeCustomer()
+    {
+        $this->assertCustomerExists();
+
+        $this->stripe()->customers->delete($this->stripe_id);
+
+        $this->stripe_id = null;
+
+        $this->save();
+    }
+
+    /**
      * Get the Stripe customer for the model.
      *
      * @param  array  $expand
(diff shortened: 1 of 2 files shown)


driesvints:
So this is actually already handled by webhooks. If you delete a customer from Stripe, the data will also get unset in the database. We actually need to keep the stripe ID on the user so all other data is unset as well: 

https://github.com/laravel/cashier-stripe/blob/14.x/src/Http/Controllers/WebhookController.php#L245

rkeshwani:
This method appears to cancel all of the user's subscriptions but maintains the customer data within Stripe, does it not?
What if a USER requests account deletion? Then their payment information/personal information is still active in Stripe as a Customer object. My PR was intended to solve that problem.

Or is there another method exposed in the codebase somewhere that already does this?

Note: My intent here was from a end user-initiated delete account perspective. NOT from a vendor initiated delete customer perspective.

rkeshwani:
Ah okay, I can solve it that way for now I guess.

PR #1529, opened 2023-04-26
[15.x] Remove isDeleted method

driesvints:
This PR removes the `isDeleted` method from the invoice object because the status doesn't exists anymore in Stripe.


driesvints:
Tests are passing btw. Just rate limiting from Stripe.

Issue #1530, opened 2023-05-05
Cannot Save Cash App as Payment Type (Webhook Error)


shatterproof:
### Cashier Stripe Version

14.12.0

### Laravel Version

9.52.7

### PHP Version

PHP 8

### Database Driver & Version

MySQL 8

### Description

Stripe Checkout recently added Cash App Pay as an option for recurring payments, however using this with Cashier causes a 500 error for the webhooks `customer.updated` and `customer.subscription.updated` with the ErrorException: `Attempt to read property "last4" on array`. This is coming from the `optional($paymentMethod)->$type->last4` code in `ManagesPaymentMethods.php`.

This happens apparently due to Cash App not providing that last4 attribute, but it is strange because the Link payment method also doesn't have the attribute but works just fine with the same webhooks.

The intended behavior is it should set the `pm_type` on the User model to `cashapp` with no error after checkout. Simply fixing this issue with the webhook should allow full support of Cashier with Cash App Pay via Stripe Checkout.

### Steps To Reproduce

Use a Stripe Checkout with Cash App Pay for a recurring subscription and the webhooks that follow will trigger the above error.

driesvints:
Hmm that's odd. I guess this payment method doesn't have that object key. I can't find an example of a full `cashapp` object. Do you have one?

PR #1531, opened 2023-05-08
[14.x] Fallback for empty pm_type object on fillPaymentMethodDetails

shatterproof:
This should fix issue https://github.com/laravel/cashier-stripe/issues/1530 by falling back to null for the `last4` attribute when it is not accessible on certain payment method types including `cashapp`, allowing support for Cash App Pay.


driesvints:
Thanks!

PR #1534, opened 2023-05-18
[14.x] Handle automatic payment method updates

shatterproof:
Vendors, including Visa and Mastercard, often reissue and update stored payment methods in place for Stripe users. This can involve updating the expiration date, but frequently also changes the entire card number as well.

The [Stripe events log](https://dashboard.stripe.com/events?type=payment_method.automatically_updated) filtered for this type of event shows how often it happens, which we are seeing several times per day in our own apps (more than I would have expected).

These automatic updates allow the payment methods to continue to be successfully charged by Stripe, but has the side effect of causing the Cashier user model to become out of sync with the outdated `pm_last_four` stored.

Fortunately, Stripe sends a `payment_method.automatically_updated` webhook when this happens that can be listened to for triggering an update of the locally stored payment method to ensure it stays in sync, which is the purpose of this PR.

Since this is a non-breaking change, I believe it is a good candidate for a 14.x release to benefit users right away. Enabling it does require adding that webhook in the Stripe dashboard though, which I can note in the upgrade guide and docs upon merge.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,13 @@
 # Upgrade Guide
 
+## Upgrading To 14.13 From 14.12
+
+### Webhook Added
+
+PR: https://github.com/laravel/cashier-stripe/pull/1534
+
+The webhook endpoint details [in your Stripe dashboard](https://dashboard.stripe.com/webhooks) should be updated to add `payment_method.automatically_updated` as one of the events sent to ensure that a user's payment method stays in sync between Cashier and Stripe when automatically updated by the issuer.
+
 ## Upgrading To 14.4 From 14.3
 
 ### Stripe API Version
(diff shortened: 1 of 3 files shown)


driesvints:
Hi @shatterproof. Although this looks good to me, I'm wondering if the `customer.updated` event already is sent alongside this to handle it? I guess not otherwise you wouldn't have sent this in?

driesvints:
Thanks @shatterproof. Can you add a section in the upgrade guide? This can target `v14.13.0`. And also feel free to send in a PR to the docs 👍

driesvints:
Thanks, feel free to mark this as ready once done.

driesvints:
As sorry, I messed that up. I changed the release in the upgrade guide instead.

driesvints:
Thanks for your work on this btw!

PR #1544, opened 2023-06-21
Remove "type" parameter for paymentMethods function

xasor:
This PR removes "type" parameter for retrieving customer payment methods because it is no longer required.

Regarding `hasPaymentMethod` and `deletePaymentMethods`, do you want that I keep with "type" argument or should I delete it too?


driesvints:
Thanks but this is a major breaking change. Right now we have no plans for this. Perhaps for 15.x

PR #1545, opened 2023-06-23
createSetupIntent add customer stripe_id option

clementmas:
Similarly to the creation of a `PaymentIntent` using the `$user->pay()` method, the `$user->createSetupIntent()` method should also create a `SetupIntent` attached to the user's Stripe Customer if it exists.

Otherwise, all Setup Intents are created as Stripe guest customers.

So this PR makes both methods for creating intents behave the same way.


taylorotwell:
Need @driesvints to review.

driesvints (approved):
LGTM!

PR #1548, opened 2023-07-04
Added expand option to Subscription latestInvoice()

quantumwebco:
Non breaking change, added expand option to the ```latestInvoice()``` method of the ```Subscription``` class
This is useful as when getting the latest invoice to check for 'requires_action' 3DS it's helpful to be able to return the payment intent. We could also have a method on the Invoice, ```requiresConfirmation()```
Invoice ```pay()``` does not throw an ```IncompletePayment``` exception, it throws a native Stripe ```CardException``` so the payment intent is not in the error object

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -1139,9 +1139,9 @@ class Subscription extends Model
      *
      * @return \Laravel\Cashier\Invoice|null
      */
-    public function latestInvoice()
+    public function latestInvoice(array $expand = [])
     {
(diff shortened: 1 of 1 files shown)


driesvints:
Unfortunately this is a breaking change to the method signature. Anyone overwriting it will see their apps break. We can do this in 15.x (master) if you want to.

quantumwebco:
OK, I see. I'll open a PR in the master branch then. Thanks

PR #1549, opened 2023-07-06
Added expand option to Subscription latestInvoice()

quantumwebco:
Breaking change PR in 14.x moved to 15.x (master)
Added expand option to latestInvoice() method so you can get the payment intent to check for 3DS requires action
When using $invoice->pay() it does not throw an Cashier IncompletePayment exception it throws a native Stripe CardException so the payment intent is not in the error object, it helps to have the payment intent on the invoice so you can check before attempting to pay()


driesvints:
Thanks!

PR #1560, opened 2023-08-30
Fix: Use constants for HTTP response status codes

localheinz:
This pull request

- [x] uses constants instead of magic numbers for HTTP response status codes


taylorotwell:
Personally find integer statuses fine.

PR #1565, opened 2023-09-07
Prevent Sync From Deleting Existing Metadata

toddstoker:
When your Billable does not overload stripeMetadata() and a billable is synced, any existing metadata on the Stripe customer is lost. 

The default behavior should be to preserve the existing metadata -- an empty array will accomplish this.

Deleting existing Stripe customer metadata is potentially devastating if you're using Stripe as a store for data that isn't persisted to your billable model.


driesvints:
Thanks @toddstoker. That's a good spot.

Issue #1570, opened 2023-09-18
Issue with product prices having multiple currencies


edulazaro:
When a user has, let's say GBP as the default user currency and the product price has USD as base currency, but also values for other currencies, including GBP, we get this error:

The currency of the selected prices (usd) does not match the customer currency (gbp).

Added the `withCurrency($currency)` and `withUserDefaultCurrency()` methods for the `SubscriptionBuilderClass` to sort it, so if the user has GBP as a currency and the product has a USD pricing which also includes a GBP option, the GBP one will be used if specified.

edulazaro:
@taylorotwell

This does not correct a bug as long as many currencies are not assigned to a price.

Providing further explanation, let's say that at Stripe we create a price and then assign an amount for each currency:

![image](https://i.ibb.co/87nMc55/currencies.jpg)

Let's assume there's a product called PlanA, it has 2 prices, one for monthly subscriptions and another for yearly, and each of them with GBP as base currency but that also with values for USD, SEK and EUR.

Now, let's say that a customer is on Sweden and already have a `stripe_id`, or in the same way, imagine the customer subscribes to the price using stripe elements directly, setting **SEK** as currency:

```php
new \Stripe\StripeClient(STRIPE_SECRET_HERE);

$stripe->subscriptions->create([
    'customer' => auth()->user()->stripe_id,
    'items' => [[
        'price' => PRICE_ID_HERE
    ]],
    'currency' => 'SEK',
    'payment_behavior' => 'default_incomplete',
    'payment_settings' => ['save_default_payment_method' => 'on_subscription'],
    'expand' => ['latest_invoice.payment_intent'],
    'metadata' => [
       'name' => HERE_THE_PLAN_A_NAME,
     ]
]);
```

This customer will have SEK as its main currency, or might have SEK assigned from before (the currency cannot be changed once subscribed).

Now, imagine that next month this customer wants to subscribe to PlanB. which also has a price in GBP as base currency but that also has a price for USD, SEK and EUR.

I will let the customer pay with the card which is already saved as the default payment method, so I will do:

```php
auth()->user()->newSubscription(PLAN_B_NAME, PRICEB_ID_HERE)->create(PM_ID);
```

This is what will happen:

![image](https://i.ibb.co/QN3W6fc/errr.jpg)

[...]

PR #1571, opened 2023-09-19
Add the withCurrency and withUserDefaultCurrency SubscriptionBuilderClass Methods

edulazaro:
Now it's possible o specify the currency for the subscriptions, so it's possible to use the customer currency with prices which have values on many currencies.

There are 2 tests included. On the second test, when a subscription is created with X currency, then X become the default currency for the customer. After that, the `withUserDefaultCurrency` method is used with another subscription, selecting the user default currency. Otherwise, the subscription will fail, as the price base currency will be selected, showing:

`The currency of the selected prices (usd) does not match the customer currency (eur).`

Related issue https://github.com/laravel/cashier-stripe/issues/1570

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -38,6 +38,13 @@ class SubscriptionBuilder
     protected $name;
 
     /**
+     * The user currency.
+     *
+     * @var string
+     */
+    protected $currency;
+
+    /**
      * The prices the customer is being subscribed to.
      *
      * @var array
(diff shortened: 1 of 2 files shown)


taylorotwell:

edulazaro:
@taylorotwell

This does not correct a bug as long as many currencies are not assigned to a price.

Providing further explanation, let's say that at Stripe we create a price and then assign an amount for each currency:

image

Let's assume there's a product called PlanA, it has 2 prices, one for monthly subscriptions and another for yearly, and each of them with GBP as base currency but that also with values for USD, SEK and EUR.

Now, let's say that a customer is on Sweden and already have a `stripe_id`, or in the same way, imagine the customer subscribes to the price using stripe elements directly, setting **SEK** as currency:

```php
$stripe->subscriptions->create([
    'customer' => auth()->user()->stripe_id,
    'items' => [[
        'price' => PRICE_ID_HERE
    ]],
    'currency' => 'SEK',
    'payment_behavior' => 'default_incomplete',
    'payment_settings' => ['save_default_payment_method' => 'on_subscription'],
    'expand' => ['latest_invoice.payment_intent'],
    'metadata' => [
       'name' => HERE_THE_PLAN_A_NAME,
     ]
]);
```

This customer will have SEK as its main currency, or might have SEK assigned from before (the currency cannot be changed once subscribed).

Now, imagine that next month this customer wants to subscribe to PlanB. which also has a price in GBP as base currency but that also has a price for USD, SEK and EUR.

I will let the customer pay with the card which is already saved as the default payment method, so I will do:

```php
auth()->user()->newSubscription(PLAN_B_NAME, PRICEB_ID_HERE)->create(PM_ID);
```

This is what will happen:

image

This is because it's not possible to specify the currency, so the default one will be used, which might no be the same as the customer currency. To sort it, it's enough with sending the currency to Stripe, so I just added the 2 methods above, so you can do:

```php
auth()->user()->newSubscription(PLAN_B_NAME, PRICE_ID_HERE)->withCurrency('SEK')->create(PM_ID);
```
The subscription will be successfully created.

Or also:

```php
auth()->user()->newSubscription(PLAN_B_NAME, PRICE_ID_HERE)->withUserDefaultCurrency()->create(PM_ID);
```
If it cannot be added, we can use a fork or extend some classes meanwhile.

Thanks for reading!

PR #1575, opened 2023-09-26
Check all subscriptions when running `$user->subscribed('name')`

maxheckel:
Currently, if a user has multiple subscriptions with the same name where one is canceled and another is active and the canceled subscription was created after the active subscription cashier will return that the user is not subscribed.  This is due to the subscription pulling just the `first()` model out of the database.  Instead we should check every subscription with the same name and return if any of them are active.

As an example, cashier was returning false for `subscribed('unlimited')` with the following database state:
image

### DIFF
--- a/src/Concerns/ManagesSubscriptions.php
+++ b/src/Concerns/ManagesSubscriptions.php
@@ -135,13 +135,14 @@ trait ManagesSubscriptions
      */
     public function subscribed($name = 'default', $price = null)
     {
-        $subscription = $this->subscription($name);
-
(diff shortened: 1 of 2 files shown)


driesvints:
The current behavior is the correct behavior. The latest created subscription is always the one you should check. All the ones before need to be canceled/handled before starting a new one. These records are purely held for historic purposes. Please check the docs: https://laravel.com/docs/10.x/billing#checking-subscription-status

Screenshot 2023-09-26 at 14 04 37

Issue #1578, opened 2023-10-02
Checkout overwrites default `subscription_data` with `$sessionOptions`


edalzell:
### Cashier Stripe Version

14.12.7

### Laravel Version

10.25.2

### PHP Version

8.2.11

### Description

I was trying to create a Checkout Session with a trial period and not collect payment information.

According to the Stripe [docs](https://stripe.com/docs/billing/subscriptions/trials#configure-free-trials-without-payment-methods-to-cancel), you need to set the payload to something like this:

```php
                'payment_method_collection' => 'if_required',
                'subscription_data' => [
                    'trial_period_days' => 14,
                    'trial_settings' => [
                        'end_behavior' => ['missing_payment_method' => 'cancel'],
                    ],
                ],
```

However, if you pass that in to the `checkout` method, the `array_merge` gets rid of the "default" payload, including:

```php
                    'metadata' => ['name' => 'default'],
```

I believe what's needed is [`array_merge_recursive`](https://www.php.net/manual/en/function.array-merge-recursive.php) instead of `array_merge` [here](https://github.com/laravel/cashier-stripe/blob/5c831f122b1e27d63862495653761421e512ff0d/src/SubscriptionBuilder.php#L364)

### Steps To Reproduce

Try to create a checkout session using:

```php
            ->checkout([
                'payment_method_collection' => 'if_required',
                'subscription_data' => [
                    'trial_period_days' => 14,
                    'trial_settings' => [
                        'end_behavior' => ['missing_payment_method' => 'cancel'],
                    ],
                ],
            ]);
```

And you'll see the metadata is missing from the Stripe request.

edalzell:
The work-around is to add the default fields:

```php
            ->checkout([
                'payment_method_collection' => 'if_required',
                'subscription_data' => [
                    'metadata' => ['name' => 'default'],
                    'trial_period_days' => 14,
                    'trial_settings' => [
                        'end_behavior' => ['missing_payment_method' => 'cancel'],
                    ],
                ],
            ])->redirect();
```

driesvints:
Hey @edalzell. Thanks for this. I believe you're right. I've sent in a PR for this: https://github.com/laravel/cashier-stripe/pull/1579

PR #1579, opened 2023-10-03
[14.x] Fix array merge for checkout

driesvints:
Makes sure the payload for checkout is recursively merged instead of overwritten. The `create` method of the builder doesn't suffers from this because the default payload doesn't have multidimensional options.

Fixes https://github.com/laravel/cashier-stripe/issues/1578

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -361,7 +361,8 @@ class SubscriptionBuilder
             ]),
         ]);
 
-        return Checkout::customer($this->owner, $this)->create([], array_merge($payload, $sessionOptions), $customerOptions);
+        return Checkout::customer($this->owner, $this)
+            ->create([], array_merge_recursive($payload, $sessionOptions), $customerOptions);
     }
 
     /**


edalzell:
Thanks @driesvints!

I was heading back here to do a PR (did the issue at the end of my day) only to find you'd done it already! You rock.

bjarn:
Just dropping this here as this PR might be a breaking change to others like it was for me:

When I first used Cashier, I had to (or at least read in Stripe or Cashier's docs) that I had to set `mode` to `subscription` in the checkout attributes.
Because of this change, the mode gets merged and changed into an array instead of string, causing this error:
CleanShot 2023-10-14 at 13 04 48

Fixed by removing the `mode` attribute from my checkout builder.

driesvints:
@bjarn can you share the code that causes this?

bjarn:
@driesvints Sure!

I used the following builder:

```php
        $builder = $team->newSubscription(
            'default',
            $plan->priceId($data->billing_cycle)
        );
        
        // ...

        return $builder->checkout([
            'mode' => 'subscription',
            'client_reference_id' => $team->id,
            'customer_update' => [
                'name' => 'auto',
                'address' => 'auto',
                'shipping' => 'never',
            ],
            'billing_address_collection' => 'required',
            'allow_promotion_codes' => true,
            'success_url' => xxx,
            'cancel_url' => xxx
        ])->asStripeCheckoutSession()->url;
```

I'm not sure why the mode ended up here, but I saw it was also set here:
https://github.com/laravel/cashier-stripe/pull/1579/files#diff-6ddbc0726400777bbee69d4448fa971720fb206f613abda0e165fffd8a340316L356

Now it is recursively merged, I think it's appended instead of overwritten of course. Probably have overlooked something when diving into the Stripe docs to see all available parameters not thinking about Cashier's abstractions.

Removed the mode from the builder and it was all fine again.

driesvints:
Yeah, I'm sorry but I don't feel we can consider this an issue since the builder already adds this for you.

PR #1586, opened 2023-10-25
Bug 1585 - Tax ID collection requires updating business name on the customer

### LINKED ISSUE #1585: Tax ID collection requires updating business name on the customer.
### Cashier Stripe Version

13.17

### Laravel Version

10.8

### PHP Version

8.1

### Database Driver & Version

_No response_

### Description

Hello everybody,

I was using Cashier with Stripe (checkout). It was working normally, but I started getting this error:

```
"Tax ID collection requires updating business name on the customer. To enable tax ID collection for an existing customer, please set `customer_update[name]` to `auto`.
```

```
 {
      "error": {
        "message": "Tax ID collection requires updating business name on the customer. To enable tax ID collection for an existing customer, please set `customer_up ▶
        "request_log_url": "https://dashboard.stripe.com/test/logs/******",
        "type": "invalid_request_error"
      }
    }
```

### Steps To Reproduce

Este é o meu ```AppServiceProvider```:

```
<?php

namespace App\Providers;
...

class AppServiceProvider extends ServiceProvider
{
    /**
     * Register any application services.
     */
    public function register(): void
    {
        //
    }

    /**
     * Bootstrap any application services.
     */
    public function boot(): void
    {
    }
}
```

I'm not using ```Cashier::calculateTaxes();```.

This is my model  ```User``` which must be billable:

```
<?php

namespace App\Models;
...

class User extends Authenticatable implements MustVerifyEmail
{
    use HasApiTokens;
    use HasFactory;
    use HasProfilePhoto;
    use Notifiable;
    use TwoFactorAuthenticatable;
    use HasRoles;
    use Billable;

    protected $fillable = [
        'name',
        'email',
        'password',
    ];

    protected $hidden = [
        'password',
        'remember_token',
        'two_factor_recovery_codes',
        'two_factor_secret',
    ];

    protected $casts = [
        'email_verified_at' => 'datetime',
    ];

}

```

And this is the controller where I try to make the subscription:

```
class SubscriptionController extends Controller
{

    public function __invoke(Request $request, Plan $plan)
    {
        try {
            /* @var User $user*/
            $user = auth()->user();
            $stripeCustomer = $user->createOrGetStripeCustomer();

            $subscriptionBuilder = $user->newSubscription('default', $plan->stripe_id);
            if ($plan->free_trial && !$user->onTrial()){
                $subscriptionBuilder->trialDays(31);
            }

            return $subscriptionBuilder->checkout([
                'success_url' => route('dashboard'),
                'cancel_url' => route('dashboard'),
            ])->redirect();
           

        }catch (\Exception $e){
            ...
        }

    }
}
```

All migrations were executed.

My .env file has the registered keys:

```
CASHIER_CURRENCY=brl
STRIPE_KEY=*******
STRIPE_SECRET=*******
STRIPE_WEBHOOK_SECRET=*******
```

stevebauman:
Closes #1585 

This PR resolves an issue that appears to have been introduced from a Stripe 2023-10-16 API update, where only the existence of the `tax_id_collection.enabled` parameter requires the addition of the `customer_update` parameters, rather than being able to set it to `false` as it was previously, resulting in the exception:

```
Stripe\Exception\InvalidRequestException

Tax ID collection requires updating business name on the customer. To enable tax ID collection for an existing customer, please set `customer_update[name]` to `auto`.

```

The `tax_id_collection` collection parameter will now be completely omitted when auto-calculation of taxes is disabled.

### DIFF
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -85,7 +85,7 @@ class Checkout implements Arrayable, Jsonable, JsonSerializable, Responsable
         }
 
         // Make sure to collect address and name when Tax ID collection is enabled...
-        if (isset($data['customer']) && $data['tax_id_collection']['enabled'] ?? false) {
+        if (isset($data['customer']) && ($data['tax_id_collection']['enabled'] ?? false)) {
             $data['customer_update']['address'] = 'auto';
             $data['customer_update']['name'] = 'auto';
         }
--- a/src/CheckoutBuilder.php
+++ b/src/CheckoutBuilder.php
@@ -79,9 +79,9 @@ class CheckoutBuilder
 
                 return $item;
             })->values()->all(),
-            'tax_id_collection' => [
-                'enabled' => Cashier::$calculatesTaxes ?: $this->collectTaxIds,
-            ],
+            'tax_id_collection' => (Cashier::$calculatesTaxes ?: $this->collectTaxIds)
+                ? ['enabled' => true]
+                : [],
         ]);
 
         return Checkout::create($this->owner, array_merge($payload, $sessionOptions), $customerOptions);


src/Checkout.php
```diff
@@ -85,7 +85,7 @@ public static function create($owner, array $sessionOptions = [], array $custome
         }
 
         // Make sure to collect address and name when Tax ID collection is enabled...
-        if (isset($data['customer']) && $data['tax_id_collection']['enabled'] ?? false) {
+        if (isset($data['customer']) && ($data['tax_id_collection']['enabled'] ?? false)) {
```
stevebauman, line 88: Discovered that this condition previously didn't evaluate correctly without parenthesis, as the null coalescing operator `??` has a lower precedence than the `&&` operator, so the condition is equivalent to:

```php
(isset($data['customer']) && $data['tax_id_collection']['enabled']) ?? false
```

driesvints:
Thanks @stevebauman!

maskeynihal:
Thanks @stevebauman for the fix.

novaylt123:
This should've been in an announcement. I wasn't aware some of my projects were broken on Stripe side of things.
But thanks for the fix team.

driesvints:
@novaylt123 I still need to do a more thorough investigation but I'm not sure why things suddenly broke at all. Cashier right now doesn't supports `2023-10-16` as it's pinned to `2022-11-15`: https://github.com/laravel/cashier-stripe/blob/14.x/src/Cashier.php#L28

So I want to know why and how this API updates affects Cashier at all.

driesvints:
So Stripe got back to me and confirmed this was a breaking change in test mode only as they were trying some things out. Nothing got broken in production and they've reverted the change since.

stevebauman:
Thanks for getting to the bottom of it @driesvints! 🙏

PR #1588, opened 2023-10-25
Update CheckoutBuilder.php

Ronan-Lenor:
regarding : #1585

in addition of : #1586

(with script style compliance)


Ronan-Lenor:
the most recent script is better than my solution given 1 hour ago

PR #1602, opened 2023-11-15
Remove invoice.payment_succeeded webhook

clementmas:
It looks like the webhook `invoice.payment_succeeded` is not used anymore.

So this PR removes it from the default list of webhooks that are registered when running:

```bash
php artisan cashier:webhook
```

See documentation: https://laravel.com/docs/10.x/billing#handling-stripe-webhooks


driesvints:
This one is used in Spark.

clementmas:
People using Cashier don't necessarily use Spark. I was trying to figure out why I was receiving this webhook. It's a bit confusing. I'll remove it manually then.

Issue #1590, opened 2023-10-25
Cashier support for Stripe 'embedded checkout'


joshembling:
Stripe has recently released a new embedded option for Stripe Checkout. Will Cashier be supporting this addition anytime soon?

https://embedcheckout.com/
https://stripe.com/docs/payments/accept-a-payment?platform=web&ui=embedded-checkout

driesvints:
Could definitely be cool. Would love to get a PR for this.

driesvints:
Gonna close this one since https://github.com/laravel/cashier-stripe/pull/1605 got merged.

PR #1605, opened 2023-11-19
[14.x] Allow creation of checkouts with ui_mode=embedded

delassiter:
It is currently not possible to create embedded checkouts with cashier-stripe (see  #1590). Creating such a checkout requires a parameter `ui_mode` to be set to `embedded`.

This fails, because Stripe does not accept a `success_url` or `cancel_url` in this mode. Since these are added by the `Checkout` class on every creation, there is no way of preventing these parameters from being sent to Stripe and causing an error.

My minimal solution only adds `success_url` and `cancel_url` if `ui_mode` is not explicitly set to `embedded`, otherwise it enforces the existence of the `return_url` parameter.

A limitation of this solution is, that if the user explicitly set the `success_url` or `cancel_url` and is using the embedded mode, an error by Stripe will still appear, since these values are not removed from the payload if incompatible.

### DIFF
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -94,7 +94,7 @@ class Checkout implements Arrayable, Jsonable, JsonSerializable, Responsable
             $data['subscription_data']['metadata']['is_on_session_checkout'] = true;
         }
 
-        // Make sure appropriate urls for ui_mode are set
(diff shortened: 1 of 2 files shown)


driesvints (approved):
Thanks @delassiter!

PR #1615, opened 2023-12-13
[15.x] Stripe API and SDK update

### LINKED ISSUE #1556: Support for Stripe PHP v12
:wave: Hi,

Just bumping the most recent [Stripe](https://github.com/stripe/stripe-php/releases/tag/v12.0.0) changes that have been released. Wondered if Cashier will try to support changes?

driesvints:
This updates the Stripe API to the most recent one `2023-10-16` as well as update the SDK version to v13 which is compatible with it.

I also decided to drop support for older SDK versions because v13 is the one that's compatible with this API version.

Closes #1556

### DIFF
--- a/src/Concerns/ManagesInvoices.php
+++ b/src/Concerns/ManagesInvoices.php
@@ -160,10 +160,17 @@ trait ManagesInvoices
     {
         $this->assertCustomerExists();
 
+        $stripeCustomer = $this->asStripeCustomer();
+
         $parameters = array_merge([
             'automatic_tax' => $this->automaticTaxPayload(),
             'customer' => $this->stripe_id,
+            'currency' => $stripeCustomer->currency ?? config('cashier.currency'),
         ], $options);
+
+        if (isset($parameters['subscription'])) {
+            unset($parameters['currency']);
+        }
 
         if (array_key_exists('subscription', $parameters)) {
             unset($parameters['pending_invoice_items_behavior']);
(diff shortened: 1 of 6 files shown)


src/Concerns/ManagesInvoices.php
```diff
@@ -160,11 +160,18 @@ public function createInvoice(array $options = [])
     {
         $this->assertCustomerExists();
 
+        $stripeCustomer = $this->asStripeCustomer();
+
         $parameters = array_merge([
             'automatic_tax' => $this->automaticTaxPayload(),
             'customer' => $this->stripe_id,
+            'currency' => $stripeCustomer->currency ?? config('cashier.currency'),
```
driesvints, line 168: There cannot be multiple currencies anymore on invoices so we'll use the one from the customer or fallback to the Cashier default.

src/Concerns/ManagesInvoices.php
```diff
@@ -160,11 +160,18 @@ public function createInvoice(array $options = [])
...
         ], $options);
 
+        if (isset($parameters['subscription'])) {
+            unset($parameters['currency']);
+        }
+
```
driesvints, line 174: `currency` cannot be combined with `subscription` so we allow `subscription` to take precedence.

src/Concerns/PerformsCharges.php
```diff
@@ -89,6 +89,10 @@ public function createPayment($amount, array $options = [])
             $options['customer'] = $this->stripe_id;
         }
 
+        if ($options['confirm'] ?? false) {
+            $options['return_url'] ??= route('home');
+        }
+
```
driesvints, line 95: When a payment intent is being confirmed, we'll always need to supply a return url.

PR #1619, opened 2023-12-18
Allow subscribed and other methods to check for any active subscription and not just one with the name "default"

joelwmale:
Currently the code always expects a subscription called "default" when checking any of the subscribed, or on trial methods without passing an argument.

For my specific use case (and I do believe many others), we have a few products all with ids `prod_xxxx` with meaningful names like `Monthly Subscription`, and `Annual Subscription`, and we only want to know if a user is subscribed to any one of these by default (i.e to hide links to subscribe, and show subscription functionality) and don't want to do multiple if checks, checking if they are subscribed to one of our specific plans.

We have already solved this on our project, by simply copying the changes in this PR into the User model, to override the Cashier functions, but to me in 2023 it makes no sense to always assume it is titled `default`.

This should not break any existing implementations, unless they have multiple subscriptions, with one called `default`. In this case, the code may or may not return the right subscription with no arguments.

### DIFF
--- a/src/Concerns/ManagesSubscriptions.php
+++ b/src/Concerns/ManagesSubscriptions.php
@@ -24,11 +24,11 @@ trait ManagesSubscriptions
     /**
      * Determine if the Stripe model is on trial.
      *
-     * @param  string  $name
+     * @param  string|null  $name
      * @param  string|null  $price
      * @return bool
      */
-    public function onTrial($name = 'default', $price = null)
+    public function onTrial($name = null, $price = null)
     {
         if (func_num_args() === 0 && $this->onGenericTrial()) {
             return true;
@@ -46,11 +46,11 @@ trait ManagesSubscriptions
     /**
      * Determine if the Stripe model's trial has ended.
      *
-     * @param  string  $name
+     * @param  string|null  $name
      * @param  string|null  $price
      * @return bool
      */
-    public function hasExpiredTrial($name = 'default', $price = null)
+    public function hasExpiredTrial($name = null, $price = null)
     {
         if (func_num_args() === 0 && $this->hasExpiredGenericTrial()) {
             return true;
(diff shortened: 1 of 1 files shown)


driesvints:
Hi @joelwmale. This unfortunately is a breaking change so we can't do this in 14.x. I'm working on 15.x and am already making a change here as well: https://github.com/laravel/cashier-stripe/pull/1620

As for your specific use case I believe you're mis-using products. You should use products as a "plan" type like a Basic, Business and/or Premium subscription. Each of these products can have "monthly" and "yearly" prices. 

I do understand a need to check if a customer is subscribed to any subscription at all. I'll try to see if I can add a few general checks to Cashier Stripe in v15

driesvints:
Actually you got me thinking and I started working on refactoring these in a similar fashion as your PR heh. Just need to crack the `trialEndsAt` method.

driesvints:
See https://github.com/laravel/cashier-stripe/pull/1621

joelwmale:
> Actually you got me thinking and I started working on refactoring these in a similar fashion as your PR heh. Just need to crack the `trialEndsAt` method.

Haha beautiful. I do see your point as well that we may be using stripe wrong, but happy to actually just have another method as an attribute that is `subscribed` if that is possible as well? 

Attribute - subscribed = yes if any active plan or trial
Method - subscribed = defaults to "default" or takes the name and checks.

But either way it all works (and also not sure if a method/attribute named the same is possible/works)!

driesvints:
Heh, I think I'd rather avoid attributes right now as well as naming two things the same but have different outcomes. I'm close to a solution just need to dig in a little more.

driesvints:
@joelwmale looks like we won't be making this change after all sorry.

joelwmale:
> @joelwmale looks like we won't be making this change after all sorry.

All good @driesvints. It is a really big breaking change and is a fundamental change to the package so I do understand Taylor's position of getting more community feedback first.

My change would only affect those who do this already because they have a product titled "default" so it is a more low risk change but nonetheless it is breaking.

I've just extended the functions to change the logic in my user class, so this issue isn't actively affecting me!

PR #1623, opened 2023-12-19
[14.x] Update Subscription.php

logcastelli:
When the Stripe customer's panel is configured to cancel immediately, the Webhook does not set the "ends_at" field, only the stripe_status is changed to "canceled". 

Therefore, the canceled() method returns false when in fact the subscription is canceled. With this change, the method also takes into account when the "stripe_status" field is canceled.

image

image

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -290,7 +290,7 @@ class Subscription extends Model
      */
     public function canceled()
     {
-        return ! is_null($this->ends_at);
+        return ! is_null($this->ends_at) || $this->stripe_status == StripeSubscription::STATUS_CANCELED;
     }
 
     /**


driesvints:
Hi @logcastelli. Trying to wrap my head around this... 

1. Could you please post a full payload of the event you mention? Is it subscription updated or deleted event?
2. Does the subscription deleted event come in at all?

logcastelli:
Hi @driesvints!

This event is received when the customer unsubscribes via Stripe's customer portal.

When the option in the portal configuration is: "Cancel at end of billing period" stripe sends an event of type "customer.subscription.updated" to the webhook.

Screenshot 2023-12-20 at 08 21 02


When the option is: "Cancel immediately" stripe sends an event of type "customer.subscription.deleted"

Screenshot 2023-12-20 at 08 21 14

As you can see in the "customer.subscription.deleted" cancel event the "cancel_at" field is null

In practice, this means that a subscription canceled on the Stripe portal configured with the "Cancel immediately" option continues to return as active and the Billable Model continues with the subscribed option equal to "true".


Full payloads below

[customer.subscription.deleted.txt](https://github.com/laravel/cashier-stripe/files/13726741/customer.subscription.deleted.txt)
[customer.subscription.updated.txt](https://github.com/laravel/cashier-stripe/files/13726742/customer.subscription.updated.txt)

* Another option would be to modify the webhook to set the ends_at field when "cancel_at_period_end": false,

driesvints:
@logcastelli but the `cancel_at` doesn't matter. The incoming `customer.subscription.deleted` event is handled here:

https://github.com/laravel/cashier-stripe/blob/0d38b0c24cc761f53dfcbc9d74b6ba40d61c8964/src/Http/Controllers/WebhookController.php#L217

Which in turn calls:

https://github.com/laravel/cashier-stripe/blob/0d38b0c24cc761f53dfcbc9d74b6ba40d61c8964/src/Subscription.php#L1072

Which will fill out the `ends_at`. The only reason I see that not happening for you is if you're not sending the event to your webhook endpoint. Could you maybe verify that you're receiving the event in your app?

Maybe I'm missing something else but the subscription should definitely be canceled.

logcastelli:
I checked here and the webhook should indeed set the ends_at field.

PR #1624, opened 2023-12-21
[14.x] Use container to resolve `StripeClient`

orkhanahmadov:
This PR uses the Laravel container to resolve `StripeClient` instead of initializing it in the Cashier class.

In complex test scenarios, this would allow us to mock the StripeClient instead of sending actual requests to Stripe and make those test cases focus only on actual business logic.


driesvints (approved):
Good idea 👍

PR #1628, opened 2024-01-05
[15.x] Clear the generic trial upon subscription creation

driesvints:
These code changes will make sure a billable's generic trial is cleared when they start their subscription. As soon as a customer starts their subscription there's no need for the trial to continue since they have started paying. We also already do this in Spark Stripe.

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -99,6 +99,11 @@ class WebhookController extends Controller
                     ]);
                 }
             }
+
+            // Terminate the billable's generic trial if it exists...
+            if (! is_null($user->trial_ends_at)) {
+                $user->update(['trial_ends_at' => null]);
+            }
         }
 
         return $this->successMethod();


AntonioDiPassio-AppSys:
Why does the generic trial need to be cleared? Would it not be better to check if the generic trial has expired, and if not set the end date to now()?
This still ensures there is no ongoing trial when a subscription is started, and at the same time you don't lose the information about when the trial ended.
Because I would still like to know from when to when my clients were on trial before they started their first subscription.

driesvints:
When a subscription starts, there's no need for a generic trial anymore because they've effectively started paying for the subscription. Previously, the customer still had the generic trial applied, making it appear as if they were still trialing, while in reality they already started the subscription.

The column in the database is really to note if a customer is currently trialing generically. After the subscription started it doesn't serves any purpose anymore. If you need this date for activity purposes we suggest you keep the info in an activity log somewhere.

AntonioDiPassio-AppSys:
Setting the end date of the trial instead of clearing it would achieve the same result: Customer isn't on trial anymore.
On top of that, it would be more accurate as the customer did end his trial the moment he started a subscription.
Clearing the end date would make it appear as if the customer never had a trial in the first place.

driesvints:
Right now we're not making any changes here. I'm sorry if this disrupted you.

PR #1643, opened 2024-01-29
Carbon v3

driesvints:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,8 +16,10 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.1, 8.2, 8.3]
-        laravel: [10, 11]
+        php: [8.2, 8.3]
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -16,8 +16,10 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.1, 8.2, 8.3]
-        laravel: [10, 11]
+        php: [8.2, 8.3]
+        laravel: [11]
+        # php: [8.1, 8.2, 8.3]
```
nunomaduro: You forgot to revert this...

nunomaduro (changes requested):

.github/workflows/tests.yml
driesvints: Done, thanks!

nunomaduro (approved):

Issue #1637, opened 2024-01-24
Cannot use proration_date for updating the quantity


skeemer:
This works for add_price() and swap() because they have $options[] that get merged.

My suggestion is to do the same thing for update_price() and all of the increment/decrement functions that pass down to it. This would be for Subscription and SubscriptionItem.

I'm willing to put together the PR if that is an acceptable solution.

driesvints:
Hi @skeemer. What do you mean with `update_price`? That method doesn't exist in Cashier.

I do agree sending along extra parameter options could be useful. But unfortunately we cannot change the method signatures on a patch release. I'm also not really comfortable polluting the public API with extra methods just to make this happen. So I guess the right option here is to target Cashier Stripe v16. If you can send in a PR we can have a look.

I btw don't know when we'd release that. For now, I suggest to copy the quantity methods into a new method you manage yourself and then swap them for the updated method whenever Cashier v16 gets released.

Thanks!

skeemer:
I meant updateQuantity(). It would just be adding the $options array parameter the same way addPrice() and swap() do it. I don't think there would be any new methods.

driesvints:
> . It would just be adding the $options array parameter 

Yeah that's what I'm trying to tell. We cannot do that. It'll be a breaking change. For example, any apps overwriting that method will break because the method signature doesn't matches anymore. It'll have to go into the next release.

PR #1648, opened 2024-02-04
Allow setting the proration_date

skeemer:
This is a different solution for #1637 that doesn't change method signatures.

This makes it easy to prorate monthly on a yearly price or any number of other splits you would like to make.

### DIFF
--- a/src/Concerns/ProratesDate.php
+++ b/src/Concerns/ProratesDate.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Laravel\Cashier\Concerns;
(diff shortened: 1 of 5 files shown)


driesvints:
Hi @skeemer. I really appreciate the effort but like I said on the other PR I don't really feel up to pollute the public API with extra methods right now. I really think an extra options parameter is the way to go here but we can do that only in the next major version. For now, I suggest to roll with your own method. Try copying the contents from the `updateQuantity` method. Sorry

Issue #1646, opened 2024-02-01
Payment behavior isn't being passed on like proration behavior is when updating the quantity from the subscription


skeemer:
### Cashier Stripe Version

14.13

### Laravel Version

10.26

### PHP Version

8.1

### Description

When calling `$someSubscription->errorIfPaymentFails()->updateQuantity(...)`, if the item is in the database, it passes the call to the SubscriptionItem without passing on the payment behavior specified.

It looks like Concerns\InteractsWithPaymentBehavior is missing setBehavior() if you match the way Concerns\Prorates works.

### Steps To Reproduce

1. Create a subscription normally and make sure the new Subscription and SubscriptionItem exist.
2. Delete the payment method within the Stripe interface.
3. `$someSubscription->errorIfPaymentFails()->updateQuantity(...)`
This will create an unpaid invoice rather than erroring.

driesvints:
Hey there,

Unfortunately we don't support this version anymore. Please check out [our support policy](https://laravel.com/docs/releases) on which versions we are currently supporting. Can you please try to upgrade to the latest version and see if your problem persists? If so, please open up a new issue and we'll help you out.

Thanks!

PR #1649, opened 2024-02-04
[15.x] Fix payment behavior not being passed

skeemer:
This fix makes sure that the requested payment_behavior is passed from Subscription to SubscriptionItem if the update is passed.

This is a fix for #1646


driesvints:
Good find! Thanks

driesvints (approved):

PR #1657, opened 2024-02-23
Allow configuring names for the `stripe_id` column on Cashier models.

kylearch:
This PR allows for using a column other than `stripe_id` on the various Cashier models.

It replaces all uses of `->stripe_id` with `->stripeId()` which can be found in the new [HasStripeId](src/Concerns/HasStripeId.php) trait. This trait is included alongside the other traits in the `Billable` trait, and in the `Subscription` and `SubscriptionItem` models. 

To make use of it, you would simply define the model's `public static function stripeIdColumn()` method to return whatever column name your model uses (including custom `Subscription` and `SubscriptionItem` models that extend their Cashier counterparts).

The purpose of this PR was so that in a project of mine I could use my custom models that mirror their actual Stripe counterparts 1:1 in my database. This allows me to both match the Stripe object schema and also more closely follow the Laravel/Eloquent schema pattern (in this case using `id` instead of `stripe_id` for the `User` [`Customer`], `Subscription`, and `SubscriptionItem` models).

### DIFF
--- a/database/factories/SubscriptionFactory.php
+++ b/database/factories/SubscriptionFactory.php
@@ -31,7 +31,7 @@ class SubscriptionFactory extends Factory
         return [
             (new $model)->getForeignKey() => ($model)::factory(),
             'type' => 'default',
-            'stripe_id' => 'sub_'.Str::random(40),
+            $this->model::$stripeIdColumn => 'sub_'.Str::random(40),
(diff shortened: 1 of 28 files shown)


driesvints:
I'm sorry but I don't think we want to take up the maintenance burden of this. If you really need this it's best that you fork the library.

PR #1660, opened 2024-02-28
Improvement: End the generic trial instead of clearing it when a subs…

AntonioDiPassio-AppSys:
This will end the generic trial instead of clearing it, which enables you to still know if a user had a generic trial or not (and when it ended)

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -100,9 +100,9 @@ class WebhookController extends Controller
                 }
             }
 
-            // Terminate the billable's generic trial if it exists...
-            if (! is_null($user->trial_ends_at)) {
-                $user->update(['trial_ends_at' => null]);
+            // End the billable's generic trial if it exists...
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not going to risk a breaking change here.

AntonioDiPassio-AppSys:
Actually, the introduction of this piece of code (in [#1628](https://github.com/laravel/cashier-stripe/pull/1628)) caused a breaking change in my application.

Regardless of that, how would my pull request be a breaking change? 
It serves the same purpose as the changed code: making sure there is no generic trial when a subscription is started. 
But it is more logical to end the ongoing generic trial at that moment instead of just clearing it (making it seam there was never a generic trial in the first place)

@driesvints Can you back me up on this one please?

driesvints:
I'm sorry @AntonioDiPassio-AppSys but looks like Taylor doesn't wants to merge this right now. Since we only got one report for this we're gonna let things be.

Issue #1102, opened 2021-03-16
Extra info from StripeSubscription


impeto:
I'm working on an app where I have to save the `current_period_start` and `current_period_end` after I create a `StripeSubscription`. Right now I do it by calling `Subscription::asStripeSubscription()` after `SubscriptionBuilder::create()` and pulling the info I need from there. However that is another call to the Stripe API and I'm thinking it could be avoided by allowing the extraction of extra info after the first API call. Perhaps a `Closure` that would take in the new `StripeSubscription` object and would return an `array`, which then can be merged with the other info before the `Subscription` is persisted to the database...?

I'm not sure what the best solution would be or even if this is needed all that often, being that this is my first project that requires it.

driesvints:
Hi there. At the moment we aren't considering adding these database columns to Cashier. I'm afraid retrieving them from Stripe is the only way. You could overwrite the Cashier models if you want (see the docs) or consider implementing caching.

impeto:
I understand, but that is not what I was asking. I already created all the extra columns I need. I was talking about a mechanism that would allow me to tap into the newly created Stripe Subscription, inside the `SubscriptionBuilder::create()` method, to avoid making an extra Stripe API call. No worries though. Thanks.

Issue #1054, opened 2021-02-05
Store subscription renewal date


JohnnyWalkerDesign:
Hi. Thanks for an awesome package. 

I understand the `ends_at` is populated when a subscription has been cancelled, but it's typical to show a user when their subscription is due for automatic renewal. This is especially helpful for yearly subscriptions where the user may have forgotten how long is left on their subscription. 

It would be really helpful if a `renews_at` column could be added.

Thanks a lot.

driesvints:
Hey @JohnnyWalkerDesign. I realise that this is probably useful but we don't intend to save every value of a Stripe subscription on the subscription's table, just the ones we need to the functionality of the package. You can retrieve this value through the `->asStripeSubscription()` method.

JohnnyWalkerDesign:
Understood. That makes sense. I will use the `->asStripeSubscription()` method.

joukhar:
which row or key refer to renews_at

Issue #874, opened 2020-02-20
Add next_payment_date to Subscription


martindrapeau:
I want to display the next payment date yet the Subscription class does not have that. As a workaround, I call `->asStripeSubscription()->current_period_end`. But that requires an API call to Stripe.

driesvints:
No plans to add this to cashier itself. Feel free to use the `current_period_end` on the strip subscription or implement this in your own app.

Issue #515, opened 2018-05-10
Need current_period_end and current_period_start column in subscriptions table


tareqtms:
After a user cancels the subscription, I need to show the user how many days remaining for the current period. There is no column for those values in the subscriptions table. 

How can I achieve this?

kyranb:
You can compare the cancels at column to the current date/time using
carbon's difference methods.

tareqtms:
Ok, that will solve my initial issue. What if I want to show the next billing date to the user (when the subscription is not canceled).

tareqtms:
Basically, I need to know, is there any way to add any additional fields to the subscriptions table, if I feel the need for any additional data?

coryrose1:
Hi @tareqtms 

I'm sure you've moved on from this issue but I found it searching for the same thing.

It's actually simpler than expected. Simply add a field to the subscriptions table using a new migration (or even manually works).

You can then update the field like so:

`
$user->subscription('main')->new_field_name = 'new_value';
$user->subscription('main')->save();
`

driesvints:
Think this issue can be closed by using the examples provided here. Feel free to provide additional feedback if you want.

stueynet:
Sorry to ressurect this but the above examples don't actually solve much as far as I can tell. `current_period_end` is a field provided by the stripe API. It is referenced in the Cashier code here: https://github.com/laravel/cashier/blob/master/src/Subscription.php#L309 however it is not accessible via the cashier subscription object which only gives you:
```
    "id" => 123
    "user_id" => 123
    "name" => "default"
    "stripe_id" => "sub_123"
    "stripe_plan" => "name_of_plan"
    "quantity" => 1
    "trial_ends_at" => null
    "ends_at" => null
    "created_at" => "2018-10-13 16:51:23"
    "updated_at" => "2018-10-13 16:57:09"
```

driesvints:
@stueynet you can do `$user->subscription()->asStripeSubscription()-> current_period_end `.

stueynet:
Oh wow thanks @driesvints thats the method I need.

yoeriboven:
@driesvints This would mean my app has to connect to the Stripe API every time a user wants to see the end of their subscription.

Using @NassimRehali15 solution means the date is incorrect until the command is run.

What's the reason for not putting this in a column on the subscriptions table? It's a value used in almost every app.

[...]

PR #1675, opened 2024-05-03
[16.x] Store subscription renewal date

clementmas:
I believe there's a real need from the community to store the subscription renewal date to display it to users or use for cron jobs. See: #361, #515, #874, #1054, #1102, [Stack Overflow](https://stackoverflow.com/questions/41576568/get-next-billing-date-from-laravel-cashier), [Laracasts](https://laracasts.com/discuss/channels/laravel/laravel-cashier-expiry-date), [gist](https://gist.github.com/garygreen/fb4dc0288e2c57f9af015968ff7019bb), etc.

The current workaround requires making an extra API request `$user->subscription()->asStripeSubscription()->current_period_end` that can be avoided if we store the renewal date `$subscription->renews_at` when creating, updating and swapping subscriptions.

Note: I considered using the `ends_at` column instead of creating a new `renews_at` column but they'll have different values when using `$subscription->cancelAt()`.

I understand the need to keep Cashier simple and functional but I think this is a relatively simple change that adds a lot of value.

This PR does not break any existing feature but it does require running this database migration to upgrade:

```php
// php artisan make:migration add_cashier_subscriptions_renews_at_column --table=subscriptions
Schema::table('subscriptions', function (Blueprint $table) {
    $table->timestamp('renews_at')->nullable()->after('quantity');
});
```

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -86,6 +86,7 @@ class WebhookController extends Controller
                     'stripe_status' => $data['status'],
                     'stripe_price' => $isSinglePrice ? $firstItem['price']['id'] : null,
                     'quantity' => $isSinglePrice && isset($firstItem['quantity']) ? $firstItem['quantity'] : null,
+                    'renews_at' => Carbon::createFromTimestamp($data['current_period_end']),
                     'trial_ends_at' => $trialEndsAt,
                     'ends_at' => null,
                 ]);
@@ -163,11 +164,14 @@ class WebhookController extends Controller
                 }
             }
 
+            // Renewal date...
+            $subscription->renews_at = Carbon::createFromTimestamp($data['current_period_end']);
+
             // Cancellation date...
             if ($data['cancel_at_period_end'] ?? false) {
                 $subscription->ends_at = $subscription->onTrial()
                     ? $subscription->trial_ends_at
-                    : Carbon::createFromTimestamp($data['current_period_end']);
+                    : $subscription->renews_at;
             } elseif (isset($data['cancel_at']) || isset($data['canceled_at'])) {
                 $subscription->ends_at = Carbon::createFromTimestamp($data['cancel_at'] ?? $data['canceled_at']);
             } else {
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -50,9 +50,10 @@ class Subscription extends Model
      * @var array
      */
     protected $casts = [
-        'ends_at' => 'datetime',
         'quantity' => 'integer',
+        'renews_at' => 'datetime',
         'trial_ends_at' => 'datetime',
+        'ends_at' => 'datetime',
     ];
 
     /**
@@ -705,6 +706,7 @@ class Subscription extends Model
             'stripe_status' => $stripeSubscription->status,
             'stripe_price' => $isSinglePrice ? $firstItem->price->id : null,
             'quantity' => $isSinglePrice ? ($firstItem->quantity ?? null) : null,
+            'renews_at' => Carbon::createFromTimestamp($stripeSubscription->current_period_end),
             'ends_at' => null,
         ])->save();
 
@@ -992,6 +994,8 @@ class Subscription extends Model
 
         $this->stripe_status = $stripeSubscription->status;
 
+        $this->renews_at = null;
+
         // If the user was on trial, we will set the grace period to end when the trial
         // would have ended. Otherwise, we'll retrieve the end of the billing period
         // period and make that the end of the grace period for this current user.
@@ -1026,6 +1030,8 @@ class Subscription extends Model
         ]);
 
         $this->stripe_status = $stripeSubscription->status;
+
+        $this->renews_at = null;
 
         $this->ends_at = Carbon::createFromTimestamp($stripeSubscription->cancel_at);
 
@@ -1078,6 +1084,7 @@ class Subscription extends Model
     {
         $this->fill([
             'stripe_status' => StripeSubscription::STATUS_CANCELED,
+            'renews_at' => null,
             'ends_at' => Carbon::now(),
         ])->save();
     }
@@ -1105,6 +1112,7 @@ class Subscription extends Model
         // no longer "canceled". Then we shall save this record in the database.
         $this->fill([
             'stripe_status' => $stripeSubscription->status,
+            'renews_at' => Carbon::createFromTimestamp($stripeSubscription->current_period_end),
             'ends_at' => null,
         ])->save();
 
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -50,10 +50,10 @@ class Subscription extends Model
      * @var array
      */
     protected $casts = [
+        'ends_at' => 'datetime',
         'quantity' => 'integer',
         'renews_at' => 'datetime',
         'trial_ends_at' => 'datetime',
-        'ends_at' => 'datetime',
     ];
 
     /**
--- a/database/factories/SubscriptionFactory.php
+++ b/database/factories/SubscriptionFactory.php
@@ -35,6 +35,7 @@ class SubscriptionFactory extends Factory
             'stripe_status' => StripeSubscription::STATUS_ACTIVE,
             'stripe_price' => null,
             'quantity' => null,
+            'renews_at' => null,
             'trial_ends_at' => null,
             'ends_at' => null,
         ];
--- a/database/migrations/2019_05_03_000002_create_subscriptions_table.php
+++ b/database/migrations/2019_05_03_000002_create_subscriptions_table.php
@@ -19,6 +19,7 @@ return new class extends Migration
             $table->string('stripe_status');
             $table->string('stripe_price')->nullable();
             $table->integer('quantity')->nullable();
+            $table->timestamp('renews_at')->nullable();
             $table->timestamp('trial_ends_at')->nullable();
             $table->timestamp('ends_at')->nullable();
             $table->timestamps();
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -309,6 +309,7 @@ class SubscriptionBuilder
             'stripe_status' => $stripeSubscription->status,
             'stripe_price' => $isSinglePrice ? $firstItem->price->id : null,
             'quantity' => $isSinglePrice ? ($firstItem->quantity ?? null) : null,
+            'renews_at' => Carbon::createFromTimestamp($stripeSubscription->current_period_end),
             'trial_ends_at' => ! $this->skipTrial ? $this->trialExpires : null,
             'ends_at' => null,
         ]);
--- a/tests/Feature/SubscriptionsTest.php
+++ b/tests/Feature/SubscriptionsTest.php
@@ -132,6 +132,7 @@ class SubscriptionsTest extends FeatureTestCase
         $this->assertFalse($user->subscription('main')->onGracePeriod());
         $this->assertTrue($user->subscription('main')->recurring());
         $this->assertFalse($user->subscription('main')->ended());
+        $this->assertSame(Carbon::today()->addMonth()->toDateString(), $user->subscription('main')->renews_at->toDateString());
 
         // Cancel Subscription
         $subscription = $user->subscription('main');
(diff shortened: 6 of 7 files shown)


src/Subscription.php
```diff
@@ -50,9 +50,10 @@ class Subscription extends Model
      * @var array
      */
     protected $casts = [
-        'ends_at' => 'datetime',
         'quantity' => 'integer',
+        'renews_at' => 'datetime',
         'trial_ends_at' => 'datetime',
+        'ends_at' => 'datetime',
```
driesvints: Can you move `ends_at` back to where it was?

driesvints (changes requested):

driesvints:
Tbh, I'd also like to see this in Cashier now.. thanks for the PR! The only downside here is that this is gonna need a new major release because current apps won't have the `renews_at` column.

sts-ryan-holton:
Is it better to start using `dateTime` instead of `timestamp` nowadays? Especially with the future timestamp issue approaching

src/Http/Controllers/WebhookController.php
```diff
@@ -163,11 +164,14 @@ protected function handleCustomerSubscriptionUpdated(array $payload)
                 }
             }
 
+            // Renewal date...
+            $subscription->renews_at = Carbon::createFromTimestamp($data['current_period_end']);
```
taylorotwell, line 168: Is this webhook endpoint invoked on every billing cycle's successful payment?

clementmas: The `customer.subscription.updated` webhook is triggered when the billing period ends, usually an hour before the renewal payment is attempted.

It is triggered again if the payment fails to set the subscription status to "past_due".

So it'll keep Cashier's `renews_at` column up to date with Stripe subscriptions' billing periods.

However, the one hour delay that might cause confusion. A user could see a future renewal date thinking the payment went through when it could actually fail a few minutes later.
A workaround would be to update `renews_at` inside another webhook instead (probably `invoice.payment_succeeded`).

<details>
<summary>These are the webhook events I recorded</summary>

Successful subscription renewal:
- `invoice.created`
- `customer.subscription.updated` (new `current_period_start|end` are set)
- [~ 1 hour delay]
- `payment_intent.created`
- `payment_intent.succeeded`
- `customer.updated`
- `invoice.updated` (status is "paid")
- `invoice.paid`
- `invoice.payment_succeeded`
- `invoice.finalized`

Failed subscription renewal:
- `invoice.created`
- `customer.subscription.updated` (new `current_period_start|end` are set)
- [~ 1 hour delay]
- `payment_intent.created`
- `payment_intent.payment_failed`
- `customer.updated`
- `invoice.updated` (status is "open")
- `customer.subscription.updated` (set subscription status to "past_due")
- `invoice.payment_failed`
- `invoice.updated`
- `invoice.finalized`

Successful subscription swap:
- `payment_intent.created`
- `payment_intent.succeeded`
- `invoiceitem.created` (proration adjustment)
- `invoice.created`
- `invoice.paid`
- `invoice.payment_succeeded`
- `invoice.finalized`
- `customer.subscription.updated`

Successful subscription creation:
- `payment_intent.created`
- `invoice.created`
- `invoice.finalized` (status is "open")
- `customer.subscription.created`
- `payment_method.attached`
- `payment_intent.succeeded`
- `invoice.updated` (status is "paid")
- `invoice.paid`
- `invoice.payment_succeeded`
</details>

driesvints: @clementmas I don't understand sorry. From your examples above, `customer.subscription.updated` is always triggered on time with the new `current_period_start|end` set? I don't see why we should listen to a different webhook? Subscription created is also covered with its own event.

clementmas: The webhooks for new and swapped subscriptions are triggered on time but there's a delay for renewed subscriptions.

The `customer.subscription.updated` webhook is triggered an hour before the payment to renew the subscription is attempted.

<details>
<summary>See webhook events with timestamps</summary>

Apr 1, 2024, 1:15:06 PM: `invoice.upcoming`
Apr 7, 2024, 1:14:27 PM: `invoice.created`
Apr 7, 2024, 1:14:27 PM: `customer.subscription.updated`
Apr 7, 2024, 2:15:18 PM: `payment_intent.created`
Apr 7, 2024, 1:15:19 PM: `payment_intent.succeeded`
...
</details>

So during that hour, the renewal date is technically wrong. It should display "today"'s date instead of next month/year.

But if that's not a concern, then yes, it's much simpler to keep relying on the `customer.subscription.created` and `customer.subscription.updated` webhooks.

driesvints: I wouldn't worry about that. I'd always rely on the `customer.subscription.updated` event.

taylorotwell: We don't feel it's confusing that it's fired an hour before the payment is attempted? What if the payment fails?

driesvints: I think it's important that we mimic exactly what Stripe has as a value on the subscription. If the renewal date is adjusted an hour before the payment is even attempted then we should take that over no matter what happens after it. If the payment fails and anything changes on the Subscription entity we'll get a new updated event with the new values.

taylorotwell: @driesvints I think we should focus more on what the user would expect to see, since customers don't care about Stripe. Do we not have way to show a truly accurate renewal date?

ybert: I think we have to change a few things here as period_ends_at from Stripe is not same as renews_at.
For example in the case of a cancellation the updated webhook is called and we should set renews_at to null instead of the end of the period.

We could do this this way :
```php
       // Period ends at
        $currentPeriodEndsAt = Carbon::createFromTimestamp($data['current_period_end']);

        // Cancellation date...
        if ($data['cancel_at_period_end'] ?? false) {
            $subscription->ends_at = $subscription->onTrial()
                ? $subscription->trial_ends_at
                : $subscription->renews_at;
            $subscription->renews_at = null;
        } elseif (isset($data['cancel_at']) || isset($data['canceled_at'])) {
            $subscription->ends_at = Carbon::createFromTimestamp($data['cancel_at'] ?? $data['canceled_at']);
            $subscription->renews_at = null;
        } else {
            $subscription->ends_at = null;
            $subscription->renews_at = $currentPeriodEndsAt;
        }
```

ybert: I think we have to change a few things here as period_ends_at from Stripe is not same as renews_at.
For example in the case of a cancellation the updated webhook is called and we should set renews_at to null instead of the end of the period.

We could do this this way :
```php
        // Period ends at
        $currentPeriodEndsAt = Carbon::createFromTimestamp($data['current_period_end']);

        // Cancellation date...
        if ($data['cancel_at_period_end'] ?? false) {
            $subscription->renews_at = null;
            $subscription->ends_at = $subscription->onTrial()
                ? $subscription->trial_ends_at
                : $currentPeriodEndsAt;
        } elseif (isset($data['cancel_at']) || isset($data['canceled_at'])) {
            $subscription->renews_at = null;
            $subscription->ends_at = Carbon::createFromTimestamp($data['cancel_at'] ?? $data['canceled_at']);
        } else {
            // Renewal date...
            $subscription->renews_at = $currentPeriodEndsAt;
            $subscription->ends_at = null;
        }
```

driesvints: @ybert I now wonder if `billing_cycle_anchor` is more what we need? https://docs.stripe.com/billing/subscriptions/billing-cycle

Although I don't know what happens for metered billing.

ybert: @driesvints I just tested it, and in my use cases, `current_period_end` and `billing_cycle_anchor` are the same after cancellation. 
I use the code pasted above in production for a while and it seems to work.

driesvints: @taylorotwell but what Stripe shows IS the accurate date at the time the event is shown. We have no way of knowing when to change the value if we delay it until the payment has been done as that's a complex process with lots of unknown and side effects that could happen. The fact that the renewal date changes when a payment fails is such a side effect. We should simple focus on reflecting what Stripe tells us. `current_period_end` will always be the correct indication when a subscription renews.

@ybert yeah I agree with the above code, feel free to adopt that in your PR.

clementmas: @driesvints we can ignore the billing period inside the `customer.subscription.updated` webhook and solely update the renewal date when receiving the `invoice.payment_succeeded` webhook.

That's the only way to really know when a subscription has been renewed.

The `invoice.payment_succeeded` webhook is triggered when swapping subscriptions, including when downgrading and the amount due is 0 (I just checked).
It's also triggered for metered billing renewals and when switching from a trial to an active subscription.

It would indeed be easier to just rely on Stripe's billing cycle but delaying the update to another webhook is not very complex. We do need to check for side effects though so I'll update the code I already have running in production to test this scenario.

Thanks @ybert for your catching the cancellation issue.

driesvints: Thanks @clementmas. I just feel there's gonna be quite a bit of side effects from relying on the payment succeeded webhook. When you're ready feel free to try this but I'd suggest to create a new draft PR so we can inspect the differences.

driesvints:
@clementmas I'm sorry but given there's no consensus on how to approach this I believe it's best to let this one go for now. Thank you for your time and work.

clementmas:
I haven't found any side effects in production from relying on the `invoice.payment_succeeded` webhook instead of `customer.subscription.updated`. The renewal date is updated properly and on time.

I updated the webhook code on my fork: https://github.com/clementmas/cashier-stripe/commit/c41246265aaf426b47e7f18070d161f30575eed1

I invite other developers to try it, improve it if necessary, and hopefully we can PR it again.

ngl5000:
@clementmas are you trying this again? It woudl be really nice to have this field available without a Stripe API call each time

clementmas:
I extracted my code into a new layer on top of Cashier to sync the `renews_at` date with my DB.

It works fine. No extra API call needed. But it could be more efficient if it was handled internally by Cashier.

<details>
<summary>Here's my implementation</summary>

```php
// App\Listeners\StripeWebhookHandledListener.php
/**
 * Handle customer.subscription.created
 *
 * Set renewal date on subscription creation
 */
protected function handleCustomerSubscriptionCreated(User $user, array $stripeSubscription)
{
    $subscription = $user->subscriptions()
        ->where(['stripe_id' => $stripeSubscription['id']])
        ->firstOrFail();

    $periodEnd = Carbon::createFromTimestamp($stripeSubscription['items']['data'][0]['current_period_end']);

    // Set renewal date
    $subscription->update(['renews_at' => $periodEnd]);
}

/**
 * Handle customer.subscription.updated
 *
 * Set or clear renewal date on subscription update
 */
protected function handleCustomerSubscriptionUpdated(User $user, array $stripeSubscription)
{
    // Ensure subscription updated event comes after created event
    // Stripe doesn't guarantee webhook delivery order, and sleep is the only reliable workaround
    sleep(1);

    $subscription = $user->subscriptions()
        ->where(['stripe_id' => $stripeSubscription['id']])
        ->firstOrFail();

    // cancel_at_period_end: set by cancel() and Stripe billing portal
    // cancel_at: set by cancelAt() with a specific date
    if (($stripeSubscription['cancel_at_period_end'] ?? false) || ($stripeSubscription['cancel_at'] ?? null)) {
        // Clear renewal date for canceled subscriptions
        $subscription->markAsNotRenewed();
    } else {
        $periodEnd = Carbon::createFromTimestamp($stripeSubscription['items']['data'][0]['current_period_end']);

        // Set renewal date for active subscriptions
        $subscription->update(['renews_at' => $periodEnd]);
    }
}

// App\Models\Subscription.php
class Subscription extends CashierSubscription
{
    public function markAsNotRenewed()
    {
        $this->update(['renews_at' => null]);
    }

    public function cancel()
    {
        $this->markAsNotRenewed();

        return parent::cancel();
    }

    public function cancelAt($endsAt)
    {
        $this->markAsNotRenewed();

        parent::cancelAt($endsAt);

        return $this;
    }

    public function markAsCanceled(): void
    {
        $this->markAsNotRenewed();

        parent::markAsCanceled();
    }

    public function resume()
    {
        $stripeSubscription = $this->updateStripeSubscription([
            'cancel_at_period_end' => false,
        ]);

        $currentPeriodEnd = $stripeSubscription->items->data[0]->current_period_end ?? null;

        $this->fill([
            'stripe_status' => $stripeSubscription->status,
            'renews_at' => $currentPeriodEnd ? Carbon::createFromTimestamp($currentPeriodEnd) : null,
            'ends_at' => null,
        ])->save();

        return $this;
    }
}

// App\Providers\AppServiceProvider.php
Cashier::useSubscriptionModel(\App\Models\Subscription::class);
```
</details>

PR #1684, opened 2024-06-27
[15.x] Check if the default payment method is valid

srmklive:
This PR adds a method **hasValidDefaultPaymentMethod** which checks whether the user has a valid default payment method stored on Stripe.

Currently in cashier, we can check whether a user has a default payment method through:

    hasDefaultPaymentMethod only checks for payment type.
    defaultPaymentMethod -- only gives the default payment method details.

This method will check:

- If **defaultPaymentMethod** returns null.
- If **defaultPaymentMethod** returns payment method details. It will verify the payment method details from Stripe.

### DIFF
--- a/src/Concerns/ManagesPaymentMethods.php
+++ b/src/Concerns/ManagesPaymentMethods.php
@@ -50,6 +50,23 @@ trait ManagesPaymentMethods
     }
 
     /**
+     * Determines if the customer currently has a valid default payment method.
+     *
+     * @return bool
+     */
+    public function hasValidDefaultPaymentMethod()
+    {
+        $paymentMethod = $this->defaultPaymentMethod();
+
+        // Return false if the payment method is null
+        if ($paymentMethod === null) {
+            return false;
+        }
+
+        return $this->resolveStripePaymentMethod($paymentMethod->id) === null;
+    }
+
+    /**
      * Determines if the customer currently has at least one payment method of an optional type.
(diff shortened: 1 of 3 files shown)


driesvints:
Bit confused here @srmklive. Why is `hasDefaultPaymentMethod` not enough for you? That should only return `true` if there's a valid default payment method in Stripe. Is that not happening for you?

srmklive:
@driesvints the ```hasDefaultPaymentMethod` method only returns if payment method type is set or not, 

```
    /**
     * Determines if the customer currently has a default payment method.
     *
     * @return bool
     */
    public function hasDefaultPaymentMethod()
    {
        return (bool) $this->pm_type;
    }
```
````It does not check whether it is a valid default payment method stored on Stripe.

driesvints:
But this type is only set if there's a valid default payment method on the customer in stripe. So again: why do you specifically need this new method?

srmklive:
We need this method is that they is no way to verify the default payment method in Cashier on Stripe. This method offers that check.

driesvints:
I still don't understand sorry. The `pm_type` method is only filled out when there's a valid default payment method. I don't see how this new method is differently other than it invokes an API call to Stripe. I'm going to close this one now but if you at some point can offer a thorough explanation why this is different from the current `hasDefaultPaymentMethod` then I'll be happy to revisit.

PR #1685, opened 2024-07-01
[15.x] Add stripe invoice mixin to invoice

olivernybroe:
When using the Invoice class, phpstan and IDE's won't be able to access any of the Stripe props, as the class is missing a mixin annotation.

### DIFF
--- a/src/Invoice.php
+++ b/src/Invoice.php
@@ -15,6 +15,9 @@ use Stripe\Customer as StripeCustomer;
 use Stripe\Invoice as StripeInvoice;
 use Symfony\Component\HttpFoundation\Response;
 
+/**
+ * @mixin \Stripe\Invoice
+ */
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @olivernybroe. There's a couple more of these sorts of wrapper classes. Could you update all of them as well?

driesvints:
Just to check: the `mixin` annotation won't also suggest method calls? Because that's not being forwarded, only properties.

olivernybroe:
@driesvints Ah, I totally missed that we were not forwarding methods also. 

Yeah then this approach won't work as it will also forward the method calls. 
Only way to do it then is to add all props as `@property` tags, but I don't think maintaining that list of props is a good approach. 

I'll might create a ticket with larastan and laravel idea instead 🤔

PR #1686, opened 2024-07-01
[15.x] Indicate if invoice is paid

driesvints:
Screenshot 2024-07-01 at 17 17 37

### DIFF
--- a/resources/views/invoice.blade.php
+++ b/resources/views/invoice.blade.php
@@ -314,7 +314,13 @@
                             <strong>Amount due</strong>
                         </td>
                         <td align="right">
-                            <strong>{{ $invoice->amountDue() }}</strong>
+                            <strong>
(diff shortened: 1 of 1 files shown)


chrysanthos:
That would be a great addition @driesvints. Should this take into account refunds/voids as well?

driesvints:
@chrysanthos no that shouldn't be part of an invoice imo.

resources/views/invoice.blade.php
```diff
@@ -49,9 +49,13 @@
 <div class="container">
     <table style="margin-left: auto; margin-right: auto;" width="100%">
         <tr valign="top">
-            <td width="160">
+            <td width="180">
```
driesvints, line 52: Sneaked this in to get the default Stripe invoice number on one line.

PR #1691, opened 2024-07-05
[15.x] Add stripe sdk headers capabilty when creating a new customer

mbarreiro85:
This allows users to use idempotency keys when creating new customers

https://docs.stripe.com/api/idempotent_requests


driesvints:
Thank you for this PR. This is perhaps the better and easiest solution indeed. However, we can't do this on a patch release as it's a breaking change to the method signatures. Also, I wouldn't call this `$headers` as that's not what these options are. They're request modifiers. So maybe `$requestOptions` are a better naming, after the Stripe object. 

Could you perhaps target 16.x with this? Thanks!

PR #1692, opened 2024-07-08
[16.x] Add requestOption when creating customers

mbarreiro85:
This allows users to use idempotency keys when creating new customers

https://docs.stripe.com/api/idempotent_requests

### DIFF
--- a/src/Concerns/ManagesCustomer.php
+++ b/src/Concerns/ManagesCustomer.php
@@ -53,11 +53,12 @@ trait ManagesCustomer
      * Create a Stripe customer for the given model.
      *
      * @param  array  $options
+     * @param  array|null  $requestOptions
      * @return \Stripe\Customer
      *
      * @throws \Laravel\Cashier\Exceptions\CustomerAlreadyCreated
      */
-    public function createAsStripeCustomer(array $options = [])
+    public function createAsStripeCustomer(array $options = [], array $requestOptions = null)
     {
         if ($this->hasStripeId()) {
             throw CustomerAlreadyCreated::exists($this);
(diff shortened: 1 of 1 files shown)


driesvints (approved):

u01jmg3:
Curious, is there a reason the default value for `$requestOptions` has been set as `null` rather than `[]` to follow suit with `$options`?

driesvints:
@u01jmg3 to match the stripe default

u01jmg3:
In that case, should `$options` (referred to as `$params` in Stripe) also have a default value of `null` because that's what Stripe uses?

https://github.com/laravel/cashier-stripe/blob/5512f24d3ff3e5f8b7aa362981bed0a75395b868/src/Concerns/ManagesCustomer.php#L61

---

- It seems odd to have a mix of default values if the convention is to follow suit with Stripe.
- I know it's Stripe's default to use `null` but `[]` does seem more fitting for a parameter that's type hinted with `array` but that's just my opinion.
- See `Customer::create()`: https://github.com/stripe/stripe-php/blob/ce69d019941cf528c036ae0d1a3a9580c7389460/lib/Customer.php#L62

```php
/**
 * Creates a new customer object.
 *
 * @param null|array $params
 * @param null|array|string $options
 *
 * @throws \Stripe\Exception\ApiErrorException if the request fails
 *
 * @return \Stripe\Customer the created resource
 */
public static function create($params = null, $options = null)
{
    // ...
}
```

driesvints:
@u01jmg3 yeah I guess you're right. Can you make that PR?

u01jmg3:
There are actually multiple places and that's just in `ManagesCustomer.php`. How would you like to proceed?

See: [`b8ababd`](https://github.com/u01jmg3/cashier-stripe/commit/b8ababd)

driesvints:
ah I actually meant the other way around and making the new param from here with a `[]` default.

PR #1697, opened 2024-07-23
Set variable's default value to be an array aligning it with other Cashier variable defaults

u01jmg3:
Original PR [#1692](https://github.com/laravel/cashier-stripe/pull/1692#issuecomment-2244370066)


driesvints:
Thanks!

PR #1704, opened 2024-08-12
[15.x] Catch missing payment intents

shatterproof:
This gracefully returns a not found exception instead of a 500 error when trying to access a non-existent or invalid payment intent ID in the browser such as from the route `/stripe/payment/xxx`.

### DIFF
--- a/src/Http/Controllers/PaymentController.php
+++ b/src/Http/Controllers/PaymentController.php
@@ -34,7 +34,7 @@ class PaymentController extends Controller
                 $id, ['expand' => ['payment_method']])
             );
         } catch (StripeInvalidRequestException $exception) {
-            abort(404, 'Payment not found');
+            abort(404, 'Payment not found.');
         }
 
         $paymentIntent = Arr::only($payment->asStripePaymentIntent()->toArray(), [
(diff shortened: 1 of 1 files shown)


shatterproof:
This was marked as a draft but should be ready to go, let me know if you need any changes @driesvints

src/Http/Controllers/PaymentController.php
```diff
@@ -25,12 +27,18 @@ public function __construct()
...
     {
-        $payment = new Payment(Cashier::stripe()->paymentIntents->retrieve(
-            $id, ['expand' => ['payment_method']])
-        );
+        try {
+            $payment = new Payment(Cashier::stripe()->paymentIntents->retrieve(
+                $id, ['expand' => ['payment_method']])
+            );
+        } catch (StripeInvalidRequestException $exception) {
+            throw new NotFoundHttpException;
```
driesvints: Can we convert this to an `abort(404, 'Payment not found');` instead?

driesvints (changes requested):

driesvints (approved):
Thanks @shatterproof

PR #1707, opened 2024-08-19
Fix failing event meter tests

Bark-fa:
This should fix the issue in the ManagesUserBilling trait, tests passed locally

### DIFF
--- a/src/Concerns/ManagesUsageBilling.php
+++ b/src/Concerns/ManagesUsageBilling.php
@@ -59,7 +59,7 @@ trait ManagesUsageBilling
     {
         $this->assertCustomerExists();
 
-        $startTime = $options['start_time'] ?? $this->created_at->timestamp;
+        $startTime = $options['start_time'] ?? 1;
 
         $endTime = $options['end_time'] ?? time();
 


driesvints:
Thanks!

driesvints:
@Bark-fa tests still fail it seems: https://github.com/laravel/cashier-stripe/actions/runs/10452831229/job/28942077001

Bark-fa:
I see it's failing on Laravel 10, is there any way I can test this locally? That's a very odd thing

driesvints:
You can use `composer require "illuminate/contracts=^10" --no-update` to install Laravel v10 components to run the test suite.

Bark-fa:
I just did, and the tests pass, is there any difference between what I would have by executing the composer command you gave me and the environment in which the tests are running? and what about the environment where the PR tests are getting evaluated on push?

driesvints:
Did you hook up your stripe testing env? See https://github.com/laravel/cashier-stripe/blob/15.x/.github/CONTRIBUTING.md

Bark-fa:
Yes I made sure of it, also ran it before I added the key just to make sure it throws a different kind of error

driesvints:
Then I don't know what's different besides the PHP versions. In any case we'll need to get this fixed or I'll have to revert all before tomorrow when we do releases.

Bark-fa:
Shot in the dark, but I'll try to call sleep(1) after reporting usage, maybe this has something to do with it, is there a way to keep the next PR open and still run the action?

driesvints:
Yeah just put it in draft and you can play around with it 👍

PR #1698, opened 2024-07-30
Add support for new Event Meters

Bark-fa:
Hello! This PR adds support for the new event meters by stripe for tracking usage in usage-based subscriptions, this is my first time contributing to cashier, and I've made every effort to maintain the style and naming conventions, and everything is fully typed, please let me know if I missed something either in the code itself or the tests.

Thank you so much for your time and effort.

PR #1708, opened 2024-08-19
Fix Failing Event Meter tests

Bark-fa:
this is a work in progress

### DIFF
--- a/src/Concerns/ManagesUsageBilling.php
+++ b/src/Concerns/ManagesUsageBilling.php
@@ -59,7 +59,7 @@ trait ManagesUsageBilling
     {
         $this->assertCustomerExists();
 
-        $startTime = $options['start_time'] ?? $this->created_at->timestamp;
+        $startTime = $options['start_time'] ?? 1;
 
         $endTime = $options['end_time'] ?? time();
 
--- a/tests/Feature/UsageBasedBillingTest.php
+++ b/tests/Feature/UsageBasedBillingTest.php
@@ -142,6 +142,7 @@ class UsageBasedBillingTest extends FeatureTestCase
         sleep(1);
 
         $user->reportMeterEvent(static::$meterEventName, 10);
+        sleep(1);
 
         $summary = $user->meterEventSummaries(static::$meterId)->first();
 
@@ -170,6 +171,7 @@ class UsageBasedBillingTest extends FeatureTestCase
         }
 
         $user->reportMeterEvent(static::$otherMeterEventName, 20);
+        sleep(1);
 
         try {
             $user->meterEventSummaries(static::$otherMeterId)->first();


Bark-fa:
@driesvints This seems to have fixed it

Bark-fa:
strange thing is, checks pass on my repo, but then once you merge it it fails

driesvints:
Gonna give this a go. Thanks!

driesvints:
Still one failing it seems: https://github.com/laravel/cashier-stripe/actions/runs/10471014163/job/28997392263
I also can't reproduce it locally..

Bark-fa:
I'm not sure why this happens only on merge, maybe there's network latency that's intensified depending on the server running the tests, which could explain why sleep() works, about the only "solution" I can think of is to add another sleep function before we query the event summary on line 105, I wish I had more insights, but at this stage is just educated guesses

driesvints:
I'll need to revert all of this unfortunately since we're out of time. Need to do releases today and we can't ship this without passing tests, I'm sorry about that. I still very much would like to get this in but we'll need to figure out why this was failing. I unfortunately don't have the time to deep dive into this.

Bark-fa:
No problem I understand, I'll keep trying, I'll send a PR with more sleep statements after the release and see if that fixes anything, if not, I'll use AWS to spin up multiple instances and try to see if I can reproduce it on any of them.

Bark-fa:
@driesvints I think I cracked it. The test was not failing because of the code; it was failing because Stripe had not had enough time to process it yet. It seems the Stripe server does not handle these requests urgently and passes them on to a queue of some sort. Given that the amount of data the test account has is huge, it slows it down further.

so when you call this  `$user->reportMeterEvent(static::$otherMeterEventName, 20);`
and then immediately try checking for usage; it won't work, if you try sometime after (up to a minute and a half in my case, though I have a slow network) it works.

I also reproduced it locally. I'm not sure how that happened; it just did. I think this is directly related to the number of meters in the account (the more meters, the slower it is to record usage), but this is just speculation on my part, it could also be that test meters are on a lower priority than production meters, again, just speculation.

To overcome this I implemented this method:

```    /**
     * Wait for a Stripe process to complete with polling
     *
     * @param callable $callback Function that returns non-null/false when condition is met
     * @param string $timeoutMessage Message to show if timeout occurs
     * @param int $timeout Maximum seconds to wait
     * @param int $interval Milliseconds between polling attempts
     * @return mixed The callback's return value
     * @throws Exception
     */
    protected function waitForStripeProcess(
        callable $callback,
        string $timeoutMessage,
        int $timeout = 180,
        int $interval = 10
    ): mixed
    {
        $start = now();

        while (now()->diffInSeconds($start) < $timeout) {
            try {
                $result = $callback();
                if ($result) {
                    return $result;
                }
            } catch (Exception $e) {
                // Continue polling on temporary failures
                report($e); // Log the error but continue trying
            }
            usleep($interval * 1000); // Convert milliseconds to microseconds
        }

        throw new Exception($timeoutMessage . " after {$timeout} seconds");
    }```

This waits for the action to be completed (a timeout of 3 minutes). Usually, on my test account, both tests are completed in 3 minutes at around a rate of 1.5 minutes each.

I use it like this:
```
        // Get summary and wait for processing
        $summary = $this->waitForStripeProcess(function() use ($user) {
            $summary = $user->meterEventSummaries(static::$otherMeterId)->first();
            return $summary && $summary->aggregated_value === 20.0 ? $summary : null;
        }, "Meter event summary was not processed");```


Unsure how to go forward in terms of a PR though, my repo is out of date now, any advice?

driesvints:
What was the original PR again? It's been a while so I don't remember 😅

Bark-fa:
it's for the stripe event meters, this one https://github.com/laravel/cashier-stripe/pull/1698

Sorry for taking this much time, I work full-time and study part-time it's been a bit hectic 😅

driesvints:
Ah thank you. At this point I'm just inclined to ship the feature without a test lol 😅 Let's just do that. Can you resend your PR?

Bark-fa:
Say no more! I'll try to have it done by today.

Bark-fa:
Done, I did not include tests in the PR as agreed but I did recreate the test file locally and ran it successfully (took 5 minutes for Stripe to register the changes for both tests), this is the file if you'd like to run it locally: https://pastebin.com/bk7jM1fj

driesvints:
Thanks @Bark-fa 🙏

PR #1724, opened 2024-11-13
[15.x] Supports PHP 8.4

crynobone:
- [x] `voku/portable-ascii`: https://github.com/voku/portable-ascii/issues/101


driesvints:
Seems the voku helper update is needed here as well @crynobone

PR #1725, opened 2024-11-17
Correct API version

jasonmccreary:
From what I can see in Stripe dev tools, the available version is `2023-08-16`. This corrects the date to align with the available versions

Assuming that's what Cashier should use, [this version](https://github.com/laravel/cashier-stripe/blob/80601df40f331c2130d519b9a54203a2b7c67f76/src/Cashier.php#L29) should be updated as well.

Screenshot 2024-11-17 at 1 22 43 PM


jasonmccreary:
Never mind. I see now these are internal versions and don't necessarily align with Stripe.

PR #1727, opened 2024-11-26
Added Support for New Usage Based Billing

Bark-fa:
This PR adds support for the [new stripe usage billing](https://docs.stripe.com/billing/subscriptions/usage-based) based on [billing meters](https://docs.stripe.com/api/billing/meter)


src/Concerns/ManagesUsageBilling.php
```diff
@@ -0,0 +1,79 @@
...
+     * @param  array  $requestOptions
+     * @return \Illuminate\Support\Collection
+     */
+    public function meterEventSummaries(string $meterId, array $options = [], array $requestOptions = []): Collection
+    {
+        $this->assertCustomerExists();
+
+        $startTime = $options['start_time'] ?? 1;
+
+        $endTime = $options['end_time'] ?? time();
```
taylorotwell: Would these almost always be passed? If so, should we just make them arguments?

src/Concerns/ManagesUsageBilling.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Get the usage records for a meter using its ID.
+     *
+     * @param  string  $meterId
+     * @param  array  $options
+     * @param  array  $requestOptions
+     * @return \Illuminate\Support\Collection
+     */
+    public function meterEventSummaries(string $meterId, array $options = [], array $requestOptions = []): Collection
```
taylorotwell: The `reportMeterEvent` uses `$meter` - this uses `$meterId`? Should they also both support Backed Enums?

src/Concerns/ManagesUsageBilling.php
```diff
@@ -0,0 +1,79 @@
...
+trait ManagesUsageBilling
+{
+    /**
+     * Get all the defined billing meters.
+     *
+     * @param  array  $options
+     * @param  array  $requestOptions
+     * @return \Illuminate\Support\Collection
+     */
+    public function meters(array $options = [], array $requestOptions = []): Collection
```
taylorotwell, line 17: If we end up supporting Enums for meter names / IDs below, should we also support them here?

src/Concerns/ManagesUsageBilling.php
Bark-fa: So, the $meter is the event name (the meter has to be configured to listen to a specific event name), but for the summaries endpoint, we need the meter ID that's created with the meter, different from the event_name. It could be a backed enum, but I envision someone using them by saving the value for their event names and meter IDs in config files or a DB if they are created dynamically.

They could be Enums, we could also have a data class that houses both the ID and the Event Name.

src/Concerns/ManagesUsageBilling.php
Bark-fa: Hi Taylor, I think so, it's required as shown here https://docs.stripe.com/api/usage-record-summary/object

So yes, I think they would be better suited as arguments to the function

src/Concerns/ManagesUsageBilling.php
Bark-fa: I think yes, if we end up passing meters around as an enum or a class, we should return them as such inside the collection.

src/Concerns/ManagesUsageBilling.php
taylorotwell: Got it - do you want to update the PR with that change?

taylorotwell:
Please mark as ready for review when the requested changes have been made.

taylorotwell:
@Bark-fa are you able to send a docs PR to laravel/docs for this? Thanks!

Bark-fa:
@taylorotwell sure, is it too much of an inconvenience if I get that done tomorrow? It's late over here 😅

driesvints:
@Bark-fa yeah just whenever you're able. No rush.

Bark-fa:
done

PR #1743, opened 2025-02-27
Fix retrieving subscription()

drsdre:
Function always returns null in error because it's not calling the function subscriptions() but a field called subscriptions.


taylorotwell:
This is intentional to not re-load them from the database if they are already loaded.

PR #1746, opened 2025-03-11
Store pm_type and pm_last_four when invoice is paid.

renatofrota:
Currently `pm_type` and `pm_last_four` remains empty when we use Stripe hosted checkout pages (even for subscriptions) as no "default payment method" is set in such cases.

I see Stripe hosted pages are becoming more and more used lately. And even for cases where I set up the payment details from my application, I consider knowing the "last (actual) payment method used" by a customer to be very relevant (more than knowing what is set as default for future payments which may never happen) for both business management reasons and customer support billing related cases, so this PR adds a new webhook handler for `invoice.payment_succeeded` event which is already listen by the webhook created with `php artisan cashier:webhook` but no handler existed yet.

For card payments, it saves card brand (with funding type) and exp. date to `pm_type` (e.g.: `Visa Credit exp. 2034-12`) + the last 4 digits to `pm_last_four`.

Note: the initial commit saves only "card" (literal) & last 4, if you prefer.

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -376,11 +376,18 @@ class WebhookController extends Controller
                 $payment->payment_method
             ));
 
-            $user->pm_type = $paymentMethod->type;
-
+            
             if ($paymentMethod->type === 'card') {
+                $card = $paymentMethod->card;
+
+                $pm_type = ucwords($card->brand) . ' '
+                    . ($card->funding === 'unknown' ? 'Credit' : ucwords($card->funding))
+                    . " exp. {$card->exp_year}-{$card->exp_month}";
+
+                $user->pm_type = $pm_type;
                 $user->pm_last_four = $paymentMethod->card->last4;
             } else {
+                $user->pm_type = $paymentMethod->type;
                 $user->pm_last_four = null;
             }
 
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -12,6 +12,7 @@ use Laravel\Cashier\Events\WebhookHandled;
 use Laravel\Cashier\Events\WebhookReceived;
 use Laravel\Cashier\Http\Middleware\VerifyWebhookSignature;
 use Laravel\Cashier\Payment;
+use Laravel\Cashier\PaymentMethod;
 use Laravel\Cashier\Subscription;
 use Stripe\Stripe;
 use Stripe\Subscription as StripeSubscription;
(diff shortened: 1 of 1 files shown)


driesvints:
Thank you for your PR. But this is not correct I'm afraid. These columns can only be filled out for the current active default payment method and we already have webhook listeners in place for those. Your use case is for payment methods which are attached to a customer but not set as the default on the customer level but rather on the subscription level. You can pull details for this by reading `default_payment_method` on the subscription itself: https://docs.stripe.com/api/subscriptions/object#subscription_object-default_payment_method

renatofrota:
Thank you for the tip, @driesvints.

Don't you foresee this being somehow supported natively on Cashier?

In case anyone needs, here's what I did to fetch `pm_type` and `pm_last_four` from the subscription's default payment method on `customer.subscription.updated` event (I don't know if there's another suggested way to hook into it):

routes/web.php
```php
Route::group([
    'prefix' => config('cashier.path'),
    'namespace' => 'App\Http\Controllers',
    'as' => 'cashier.',
], function () {
    Route::post('webhook', 'StripeWebhookController@handleWebhook')->name('webhook');
});
```

app/Http/Controllers/StripeWebhookController.php
```php
<?php

namespace App\Http\Controllers;

use Laravel\Cashier\Http\Controllers\WebhookController as CashierWebhookController;
use Laravel\Cashier\Payment;
use Laravel\Cashier\PaymentMethod;

class StripeWebhookController extends CashierWebhookController
{
    /**
     * Handle customer subscription updated.
     *
     * @param  array  $payload
     * @return \Symfony\Component\HttpFoundation\Response
     */
    protected function handleCustomerSubscriptionUpdated(array $payload)
    {
        $user = $this->getUserByStripeId($payload['data']['object']['customer']);

        if ($user) {
            $paymentMethodId = $payload['data']['object']['default_payment_method'] ?? null;

            if ($paymentMethodId) {
                $paymentMethod = new PaymentMethod($user, $user->stripe()->paymentMethods->retrieve(
                    $paymentMethodId
                ));

                if ($paymentMethod) {
                    if ($paymentMethod->type === 'card') {
                        $card = $paymentMethod->card;
            
                        $pm_type = ucwords($card->brand) . ' '
                            . __(($card->funding === 'unknown' ? 'Credit' : ucwords($card->funding)))
                            . " - " . str_pad($card->exp_month, 2, '0', STR_PAD_LEFT) . "/" . substr($card->exp_year, -2);
            
                        $user->pm_type = $pm_type;
                        $user->pm_last_four = $paymentMethod->card->last4;
                    } else {
                        $user->pm_type = $paymentMethod->type;
                        $user->pm_last_four = null;
                    }
            
                    $user->save();
                }
            }

            return parent::handleCustomerSubscriptionUpdated($payload);
        }
    }
}
```

driesvints:
I've thought about adding them to the subscription record in the past but at this point we don't want to add more columns to these tables sorry. What we can do however, is retrieve the payment method on the subscription through a new method. It would involve loading in the stripe object though.

PR #1749, opened 2025-03-25
Fix: Ensure trial_ends_at is properly parsed before checking if it's …

YonkoSam:
This update fixes an issue where calling isFuture() on trial_ends_at could cause an error if the value was stored as a string. Now, trial_ends_at is explicitly parsed with Carbon::parse() before checking if it’s in the future, ensuring proper type handling and preventing potential crashes.

### DIFF
--- a/src/Concerns/ManagesSubscriptions.php
+++ b/src/Concerns/ManagesSubscriptions.php
@@ -72,7 +72,7 @@ trait ManagesSubscriptions
      */
     public function onGenericTrial()
     {
-        return $this->trial_ends_at && $this->trial_ends_at->isFuture();
+        return $this->trial_ends_at && Carbon::parse($this->trial_ends_at)->isFuture();
     }
 
     /**


clementmas:
In which case could `$subscription->trial_ends_at` be a string?

The `Subscription.php` model has a datetime cast, so I don't see how it could be a string:

```php
protected $casts = [
    // ...
    'trial_ends_at' => 'datetime',
];
```

Edit: I was thinking about the wrong model. `$user->trial_ends_at` does require a manual cast.

YonkoSam:
i just had this problem this how i started the trial  

```php
   /**
     * Start a free trial for the user.
     */
    public function startTrial(): void
    {
        if (!$this->trial_ends_at) {
            $this->trial_ends_at = now()->addDays(2);
            $this->save();
        }
    }

driesvints:
Thanks but please see the docs: https://laravel.com/docs/12.x/billing#without-payment-method-up-front

Scherm­afbeelding 2025-03-25 om 16 55 51

YonkoSam:
this fix makes sure this don't happen even if you didn't see this warning like i did

PR #1757, opened 2025-04-30
fix(subscription_builder): add array_merge in metadata when build payload

K1000o-Lp:
The subscription type is not taken into account when calling the `buildPayload` method. This generates a problem in the builder method `create`.


taylorotwell:
What problem?

K1000o-Lp:
When creating a subscription with the builder's `create` method, it does not take into account the `type` attribute passed in the SubscriptionBuilder.

> What problem?

taylorotwell:
But what actually goes wrong? I thought `metadata` was just for arbitrary user data you want to store.

K1000o-Lp:
I noticed the error. I had the subscription creation process inside a DB transaction and hadn't realized it.

Issue #1759, opened 2025-05-20
Duplicate subscriptions and no try/catch for unique Stripe ID


MattRogowski:
### Cashier Stripe Version

15.3.2

### Laravel Version

10.48.12

### PHP Version

8.2

### Database Driver & Version

10.11.7-MariaDB-1:10.11.7+maria~ubu2204

### Description

This is a continuation of closed tickets #1734 #1678 #1307 and probably others.

As several people have stated in the comments on these tickets the duplicate subscription error is definitely a problem - we have had 20+ in the last few days. I'm actually surprised the webhook is so fast that it is received before Cashier creates the subscription in the database, but I concur with other users reporting this that it does indeed happen.

I believe a potential cause of this is the index was added in [this commit](https://github.com/laravel/cashier-stripe/commit/2fc19633d55e9c0c3eb6cf0404cbdd76ccf7a40d) 3 and a half years ago, but the upgrade guide doesn't make any mention of this, which is likely why some users do not have this index (myself included). Manually adding the index also introduces the issue of having to clear our the duplicate rows.

The second issue is that the webhook doesn't wrap the insert queries in a `try/catch` block, meaning if the index _does_ exist it will throw a database exception. Stripe will receive a 500 error response and will re-try it - the second attempt will work as either the query in the webhook handler or the create subscription function will find the existing subscription, however if Stripe receives too many non-2xx response codes it can disable the webhook, which is not desirable and should be avoided.

In summary, as well as the upgrade guide making a note to add this index, the webhook should suppress any exceptions regarding uniqueness to prevent Stripe disabling the webhook due to high error rates.

### Steps To Reproduce

As others mentioned this issue does not always occur, but if the timing is right, you will either end up with two subscriptions or a database exception.

driesvints:
Thanks @MattRogowski. Finally a good analysis about the issue 😅 

The index is correct because there should never be two subscription records with the same stripe ID. 

Can you send in a PR for both the upgrade guide as well as that try/catch approach? I'll have a look at it then.

clementmas:
FYI, I pulled the current `master-dev` branch into my production app to test the PR #1761 and I'm still seeing many duplicate subscriptions created at the exact same time. One is "incomplete" and the other one active.

What's the logic behind the creation of subscriptions in both `customer.subscription.created` and `customer.subscription.updated` event listeners?

I remember seeing a `sleep()` workaround. It's not pretty but it would be better than the current duplication problem.

[...]

Issue #1678, opened 2024-05-27
Duplicate entry stripe_id


sebastianro92:
### Cashier Stripe Version

15.0

### Laravel Version

10

### PHP Version

8.3.6

### Database Driver & Version

MYSQL, 8.0.36

### Description

I use Stripe Webhooks, but when i cancelling a customer subscription and create a new subscription, received then next error:

[2024-05-27 00:05:04] local.ERROR: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'sub_1PKunxFd5Q4HlP154MBNJXW5' for key 'subscriptions.subscriptions_stripe_id_unique' (Connection: mysql, SQL: insert into `subscriptions` (`type`, `stripe_id`, `stripe_status`, `stripe_price`, `quantity`, `trial_ends_at`, `ends_at`, `user_id`, `updated_at`, `created_at`) values (default, sub_1PKunxFd5Q4HlP154MBNJXW5, incomplete, price_1Lu0DCFd5Q4HlP15qpPSBDWb, 1, ?, ?, 1, 2024-05-27 00:05:04, 2024-05-27 00:05:04)) {"exception":"[object] (Illuminate\\Database\\UniqueConstraintViolationException(code: 23000): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'sub_1PKunxFd5Q4HlP154MBNJXW5' for key 'subscriptions.subscriptions_stripe_id_unique' (Connection: mysql, SQL: insert into `subscriptions` (`type`, `stripe_id`, `stripe_status`, `stripe_price`, `quantity`, `trial_ends_at`, `ends_at`, `user_id`, `updated_at`, `created_at`) values (default, sub_1PKunxFd5Q4HlP154MBNJXW5, incomplete, price_1Lu0DCFd5Q4HlP15qpPSBDWb, 1, ?, ?, 1, 2024-05-27 00:05:04, 2024-05-27 00:05:04)) at /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php:824)
[stacktrace]
#0 /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php(783): Illuminate\\Database\\Connection->runQueryCallback()
#1 /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php(576): Illuminate\\Database\\Connection->run()
[...]

PR #1761, opened 2025-05-24
Prevent duplicate subscription create webhookController

ishannz:
This will fix the issues that were found on 
- https://github.com/laravel/cashier-stripe/issues/1678 
- https://github.com/laravel/cashier-stripe/issues/1759

This will utilise updateOrCreate for both subscriptions and items in the handleCustomerSubscriptionCreated hook, which is invoked after handleCustomerSubscriptionUpdated, occasionally triggered by Stripe. This will resolve the duplicate entry for 'sub_xxxxxx' that causes the'subscriptions.subscriptions_stripe_id_unique' Mysql error.


driesvints:
We can try this but I doubt it'll solve every edge case. The query could be run while the updated webhook has just created the subscription record.

Can you remove the .idea dir and enable this for review again?

clementmas:
Thanks @ishannz for the PR. Can you update it so it can get merged?

I have so many duplicate subscriptions it's becoming really difficult to manage.

driesvints:
@clementmas if you send in your own PR with the changes then I can merge that one.

ishannz:
@driesvints done

driesvints (approved):

OliverWich:
Thanks for this fix, I hope this does finally solve the rare race-condition issue described in #1678, #1759 and others!

When will there be a new release of laravel/cashier with this PR in it available? We would need this fix quite urgently :D

PR #1781, opened 2025-10-06
Allow for multiple coupons on subscription

Arkitecht:
Currently you can only attach one coupon when creating a new subscription, even though the Stripe API allows for multiple coupons. 

Currently you would have to do something like this:
```
$subscriptionOptions = [
    'discounts'    => [
        ['coupon' => 'coupon1'],
        ['coupon' => 'coupon2'],
    ]
];

$subscription = $subscriptionBuilder->create('pm_card_visa', [], $subscriptionOptions);
```
instead of using the fluent `withCoupon()` method.

This update allows multiple calls to the `withCoupon()` method and sends all coupons to the subscription builder upon creation. Like so:

```
$user->newSubscription('main', static::$priceId)
    ->withCoupon('coupon1')
    ->withCoupon('coupon2')
    ->create('pm_card_visa');
```

It gracefully handles `Checkout` which still only allows for the single coupon.

### DIFF
--- a/src/Concerns/AllowsCoupons.php
+++ b/src/Concerns/AllowsCoupons.php
@@ -31,6 +31,27 @@ trait AllowsCoupons
     public bool $allowPromotionCodes = false;
 
     /**
+     * The coupon IDs being applied.
+     *
+     * @var array|null
+     */
+    public ?array $coupons = null;
+
+    /**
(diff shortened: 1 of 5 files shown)


taylorotwell:

Issue #1791, opened 2025-10-22
Active and Canceled both return true


Ronan-Lenor:
### Cashier Stripe Version

15.7.1

### Laravel Version

11.46.1

### PHP Version

8.4.1

### Description

that's because when i cancel a test clock, stripe cancel everything. so yeah the subscription don't exists anymore on Stripe and a webhook update the status of the subscription to "canceled". Upto here everything is fine. EXCEPT none of the methods used in valid() do care of that status. not even the canceled() itself 😅

So please, add something that check the status in all the method called by the valid() method.

### Steps To Reproduce

server : subscribe something.
stripe : advance a test_clock
stripe : cancel the test_clock
server : retrieve the subscription

Ronan-Lenor:
also related : https://github.com/laravel/cashier-stripe/issues/1550

krishnamitra03:
Hi, I’d like to work on this issue if it’s still open.
I’ve read through the discussion and will start by checking the Stripe webhook logic.
Please let me know if anyone else is already working on it. Thanks!

Ronan-Lenor:
Things to know about this problem.

When a `test_clock` is ongoing, Cashier don't sync the time. Cashier, which rely on `now()`, become desynchronized.

To affect the entire laravel project (all vendors) that rely on `Carbon`, the usage of `Carbon::hasTestNow()` & `Carbon::setTestNow()` become necessary. (don't forget the webhooks)

Cashier also rely on the `ends_at` attribute to guess the state of the subscription. (without looking at the `stripe_status` attribute).
aside of `canceled()` there's a method, closer to what we should expect from that last one, called `ended()`

impruthvi:
Still happens on 16.8.0.

It might be worth separating two paths here, because I think only one of them produces the "both return true" in the title.

#1850 assumed status: canceled arriving with no cancel timestamps. In that case WebhookController falls through to ends_at = null, so you get canceled() false, ended() false, active() true. Active is wrong, but canceled() is false, so not "both".

The other path is cancel_at_period_end. Stripe leaves that true on a canceled subscription - the docs describe it as whether the sub "will (if status=active) or did (if status=canceled) cancel at the end of the current billing period". WebhookController takes that branch first and sets ends_at = currentPeriodEnd(). If the cancellation happened early that's a future date, so onGracePeriod() is true, ended() is false, and you end up with stripe_status canceled, canceled() true and active() true at the same time. That one matches the title.

[...]

PR #1794, opened 2025-10-27
cancel() return true only in the right circumptance

Ronan-Lenor:
Also changed valid() to make the canceled status a priority over active() OnGracePeriod() & onTrial()

https://github.com/laravel/cashier-stripe/issues/1791

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -176,7 +176,7 @@ class Subscription extends Model
      */
     public function valid(): bool
     {
-        return $this->active() || $this->onTrial() || $this->onGracePeriod();
+        return !$this->canceled() && ($this->active() || $this->onTrial() || $this->onGracePeriod());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
I am closing this pull request because it lacks sufficient explanation, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Feel free to re-submit your change with a **thorough** explanation of the feature and tests - integration tests are preferred over unit tests. Please include it's benefit to end users; the reasons it does not break any existing features; how it makes building web applications easier, etc.

Thanks!

Ronan-Lenor:
yeah maybe i went too far about that Valid() function but how about the canceled one ?

right now it just check that "ends_at"is not null, so even if it's set to be canceled in the future, the function will return true, assuming that the future is now.

a quick fix would be at least to use also the function `isFuture()` like it has been done in many other functions

EDIT : oh i just figured out that there's a `ended()` function that do what i was expecting from the function `canceled()`
Nonetheless, i still think that either the `canceled()` or the `ended()` (or both) functions shoud check the `stripe_status` attribute

Issue #1550, opened 2023-07-06
Subscription considered as valid while canceled


axeloz:
### Cashier Stripe Version

13.17

### Laravel Version

9

### PHP Version

8.2

### Database Driver & Version

MySql

### Description

When canceling immediatly a subscription with a trial period, the subscription is still considered as valid by Cashier. 

It might be a bug although I suspect this is an expected behavior. 
In any case, that needs discussion because the subscription **is** deleted from Stripe and canceled into the database. 
So trial period is canceled as well.  Should be at least.

### 1/ Create a subscription 
```
$user
->newSubscription('default', $price->uid)
->trialUntil(new Carbon('tomorrow'))
->add()
;
```
Subscription is properly created on Stripe's side

### 2/ Delete the subscription
```
$user->subscription('default')->noProrate()->cancelNow();
```
Subscription has status "canceled" in database.
Subscription is properly deleted from Stripe user's account.

### 3/ Test the subscription
```
$this->subscribed('default') // returns true
```

Code that checks the subscription status. 
The issue comes from `$this->onTrial()`

```
    public function valid()
    {
        return $this->active() || $this->onTrial() || $this->onGracePeriod();
    }
```
in vendor/laravel/cashier/src/Subscription.php

PR #1798, opened 2025-11-03
Refactor Subscription model to properly handle canceled status

Diddyy:
## Summary
This pull request refines the subscription status logic to ensure that the `stripe_status` field is treated as the source of truth when determining whether a subscription is active, canceled, or within a grace or trial period.

Currently, subscriptions canceled directly via Stripe (for example, when a test clock is canceled or when a subscription is immediately canceled during a trial) may still appear as valid within the application. This happens because the `stripe_status` field isn't consistently considered in status validation methods.

## Changes
- `canceled()` — Now checks `stripe_status` before other attributes (`ends_at`) for clearer intent and correctness.
- `onTrial()` / `onGracePeriod()` — Refactored to use early returns for readability and to align with Laravel’s style conventions.
- Scope Methods — Updated to mirror the instance methods for consistent behavior between query and model contexts.
- Tests — Added dedicated tests covering:
  - Subscriptions canceled via Stripe’s API or test clocks.
  - Immediate cancellations during trial periods.
  - Edge cases around `ends_at` and `trial_ends_at`.

This change ensures that:
- The subscription status reflects the true state in Stripe.
- Users no longer retain access after cancellation.
- Trial and grace periods behave consistently even when canceled early.
- Code readability and maintainability are improved with clearer control flow.

All existing behavior for active and scheduled-cancellation subscriptions remains unchanged.

## Related Issues
#1550 #1791

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+.idea
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -229,6 +229,7 @@ class Subscription extends Model
     public function active(): bool
     {
         return ! $this->ended() &&
+            $this->stripe_status !== StripeSubscription::STATUS_CANCELED &&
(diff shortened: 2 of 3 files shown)


taylorotwell:

Issue #1805, opened 2025-11-28
Subscribed call only checks most recent subscription entry in the database


thebetar:
### Cashier Stripe Version

15.0

### Laravel Version

10.10

### PHP Version

8.2

### Description

In our system we manage multiple subscription where users can add a subscription of the same kind which adds another X devices to their account. When using the `subscribed` function defined in the trait `ManagesSubscriptions` it only check the validity of the subscription of a certain type that was created most recently due to the `->first()` call.
To me it seems like this should check if there is a subscription of that type for that user that is valid and not just the most recent subscription entry of that type.

Maybe I am wrong in assuming that it should work like this but currently if a user has an active subscription and adds a new subscription but during payment decides to just stick to a single subscription their prior subscription will not be checked when using the `->subscribed` method. I would love to create a fix for this so please let me know if this change in logic would be preferable :-)

Pull request can be found here: https://github.com/laravel/cashier-stripe/pull/1806

### Steps To Reproduce

1. Create an active subscription
2. Create a newer subscription of the same type that is cancelled
3. Try to get the status of a subscription using `->subscribed($type)`

thebetar:
Pull request was closed, so I assume that the maintainers do not want to add this feature.

PR #1806, opened 2025-11-28
Feature/subscription subscribed any support

thebetar:
This pull request is related to the issue I have opened here: https://github.com/laravel/cashier-stripe/issues/1805

To summarise I found out that when working with multiple subscription of the same type the `subscribed` functions within the train `ManagesSubscriptions` does not work as intended. It might be that I am the only one or the  one of a small group of people using it like this so I have created an extra parameter to solve this. The default behaviour should stay the same while also solving the issue for users experiencing the same challenge.

I think this issue only plays for platforms that allow for a user entity to have multiple of the same subscriptions for instance when an organisation has multiple subscriptions where there are scales of a subscription allows 10 devices to be connected and the more expensive 50. If the user requires only 20 devices it might be that two basic subscriptions are taken.

### DIFF
--- a/src/Concerns/ManagesSubscriptions.php
+++ b/src/Concerns/ManagesSubscriptions.php
@@ -139,11 +139,11 @@ trait ManagesSubscriptions
     {
         if (! $any) {
             $subscription = $this->subscription($type);
-    
+
             if (! $subscription || ! $subscription->valid()) {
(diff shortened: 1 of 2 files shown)


taylorotwell:
I would probably just define your own method on the same model that uses this trait. 👍

Issue #1810, opened 2025-12-01
Migrations fail for new installs


yoeriboven:
### Cashier Stripe Version

16.0

### Laravel Version

12.40.2

### PHP Version

8.4.14

### Database Driver & Version

MySQL

### Description

When you run the migrations they will be executed alphabetically like this:

```
2025_12_01_101217_add_meter_event_name_to_subscription_items_table.php
2025_12_01_101217_add_meter_id_to_subscription_items_table.php
2025_12_01_101217_create_customer_columns.php
2025_12_01_101217_create_subscription_items_table.php
2025_12_01_101217_create_subscriptions_table.php
```

This will fail because in the first migration it expects the `subscription_items` table to exist which at that point has not been created yet. 

Migrations in the package are currently sorted by the timestamp but that is overwritten when you publish.

### Possible solution
Add a prefix to the migration names like:

```
2019_05_03_000001_1_create_customer_columns.php
```

### Steps To Reproduce

Create a new laravel app
Install Cashier
Publish migrations
Run migrations

shankhadevpadam:
I got the same issue with Laravel Cashier in a fresh Laravel installation. The migration publish command is not publishing migration files from the migration folder of the package. Instead, it generate migration name with the current timestamp. As mention @yoeriboven

JaiveerChavda:
Yeah, faced the same issue yesterday.

PR #1818, opened 2026-01-25
Fix Cashier migration order for new installations

JaiveerChavda:
This PR fixes an issue where Cashier’s published migrations can run in the wrong order on fresh installations, causing child or alter migrations to execute before the base tables exist and resulting in migration failures.

On a new project, after publishing Cashier migrations, their filenames may be sorted alphabetically in a way that breaks the intended execution order. This behavior is documented in the following issue:

- https://github.com/laravel/cashier-stripe/issues/1810

To ensure deterministic and correct ordering, this change prefixes the published migration filenames with explicit sequence markers (`_1`, `_2`, `_3`, …). This guarantees that:

- Base tables are created before dependent or altering migrations.
- Fresh installs migrate cleanly.
- Existing projects are not affected, as their core tables already exist.

This aligns Cashier’s migration behavior with developer expectations and prevents new users from encountering setup failures.


taylorotwell:
Nothing fails for me on a fresh Laravel project, installing Cashier, publishing migrations, and then running them.

Issue #1819, opened 2026-02-09
Error removing price on flexible subscription


Arkitecht:
### Cashier Stripe Version

16.2

### Laravel Version

11.48

### PHP Version

8.3.21

### Description

When trying to remove a metered price from a subscription that is using the new flexible billing mode, it throws the following error:

`With `billing_mode.type=flexible` set on a subscription, the following parameters are not supported: clear_usage.`

### Steps To Reproduce

1. Create a subscription in stripe, ensuring the billing type is the newer flexible type
2. Add multiple products/prices, including one metered price
3. Use cashiers `removePrice` method to remove the metered price

hezeclark:
This is a tricky edge case with flexible billing and metered prices.

When we were building payment recovery automation, we saw similar issues with subscription modifications:

1. **Flexible billing changes how prices are tracked** - Metered usage is billed in arrears, so removing a price mid-cycle can cause validation errors

2. **Stripe requires at least one active price** - You can't remove the last price from an active subscription

Our workaround:

```php
// Before removing a price, ensure there's a replacement
$user->subscription('default')
    ->removePrice('price_old', [
        'proration_behavior' => 'create_prorations',
        'billing_cycle_anchor' => 'now'
    ])
    ->addPrice('price_new'); // Add new price first
```

If you're hitting this error during a migration or cleanup, try:

1. Pause the subscription first
2. Remove the problematic price
3. Resume with the new configuration

We built RecoverKit (https://tryrecoverkit.com?utm_source=github&utm_medium=issue_comment&utm_campaign=cashier_flexible34) to handle subscription management edge cases. Flexible billing adds complexity that Cashier doesn't fully abstract away.

This is a known limitation in the current implementation.

crynobone:
Fixed via #1772

PR #1772, opened 2025-08-21
[17.x] Flexible Billing Support for Laravel Cashier

Diddyy:
This PR introduces comprehensive **Flexible Billing** support to Laravel Cashier, enabling modern usage-based billing models, billing credits, and advanced subscription management that aligns with Stripe's latest Basil API capabilities.

## **What is Flexible Billing?**

Flexible Billing represents the future of subscription commerce, moving beyond traditional fixed-price subscriptions to support:

- **Usage-Based Billing**: Charge customers based on actual usage (API calls, storage, bandwidth)
- **Hybrid Models**: Combine fixed subscriptions with usage overages
- **Billing Credits**: Issue credits, refunds, and promotional balances
- **Complex Pricing**: Tiered pricing, volume discounts, and custom billing cycles
- **Advanced Scheduling**: Sophisticated subscription lifecycle management

#### **1. Usage-Based Billing (`ManagesUsageBilling` trait)**
- **Usage Thresholds**: Set, retrieve, remove, and check thresholds with overage calculations
- **Usage Analytics**: Get aggregated usage data with flexible time ranges
- **Meter Event Reporting**: Enhanced meter event tracking with improved caching
- **Percentage & Overage Calculations**: Built-in utilities for usage-based billing logic

```php
// Flexible Billing: Set usage thresholds for API-based pricing
$user->setUsageThreshold('api_calls', 10000, ['alert_email' => 'billing@company.com']);
$user->setUsageThreshold('storage_gb', 100, ['webhook_url' => 'https://app.com/webhook']);

// Calculate overages for hybrid billing models
$apiOverage = $user->calculateUsageOverage('api_calls', 15000); // Returns 5000 calls
$storageOverage = $user->calculateUsageOverage('storage_gb', 150); // Returns 50 GB

// Analytics for usage-based insights
$analytics = $user->getUsageAnalytics('api_calls', startTime: now()->subMonth());
// Returns: ['total_usage' => 45000, 'average_daily' => 1500, 'peak_day' => 2500]
```

#### **2. Billing Credits Management (`ManagesBillingCredits` trait)**
- **Customer Balance Transactions**: Create, retrieve, and manage customer credits
- **Credit Application**: Apply credits to customer accounts
- **Balance Tracking**: Enhanced credit balance monitoring

```php
// Flexible Billing: Issue promotional credits for usage-based services
$user->addBillingCredit(5000, 'New customer - 5,000 free API calls'); // $50.00 credit

// Support complex billing with account credits
[...]

PR #1820, opened 2026-02-09
Avoids Stripe error removing metered price from flex sub

Arkitecht:
This addresses the issue raised in Issue #1819 and referenced in PR #1772 [comment here](https://github.com/laravel/cashier-stripe/pull/1772#issuecomment-3488337549) until PR is merged or alternative support for flexible billing is provided.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -1023,7 +1023,7 @@ class Subscription extends Model
         $stripeItem = $this->findItemOrFail($price)->asStripeSubscriptionItem();
 
         $stripeItem->delete(array_filter([
-            'clear_usage' => $stripeItem->price->recurring->usage_type === 'metered' && !$this->usesFlexibleBilling() ? true : null,
+            'clear_usage' => $stripeItem->price->recurring->usage_type === 'metered' && ! $this->usesFlexibleBilling() ? true : null,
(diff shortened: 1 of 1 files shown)


src/Subscription.php
```diff
@@ -830,14 +830,16 @@ protected function parseSwapPrices(array $prices): Collection
...
         /** @var \Stripe\SubscriptionItem $stripeSubscriptionItem */
-        foreach ($this->asStripeSubscription()->items->data as $stripeSubscriptionItem) {
+        foreach ($stripeSubscription->items->data as $stripeSubscriptionItem) {
             $price = $stripeSubscriptionItem->price;
 
             if (! $item = $items->get($price->id, [])) {
                 $item['deleted'] = true;
 
-                if ($price->recurring->usage_type == 'metered') {
+                if ($price->recurring->usage_type == 'metered' && $this->usesFlexibleBilling($stripeSubscription)) {
```
vrymel, line 842: I think this condition is wrong. What’s expected is that when the billing mode is flexible, `clear_usage` should not be used. However, this condition is the reverse of that.

With this change, the operation would still fail when calling `swapAndInvoice` for example.

Arkitecht: Good catch, thats my bad, should be `! $this->` ....

PR #1838, opened 2026-03-30
[15.x] Supports Laravel 13

crynobone:


sgilberg:
@crynobone or @taylorotwell Can I request a 15.x release that includes this commit? I had noticed this merged commit and thought we were in the clear to upgrade to 13.x while keeping our Cashier at 15.x but it looks like the latest is 15.7.1 which is several commits behind.

The 'basil' API version that 16 is tied to has [completely dropped support for "forever" coupons](https://docs.stripe.com/changelog/basil/2025-03-31/restrict-coupon-duration), as noted also in Cashier's [upgrade guide](https://github.com/laravel/cashier-stripe/blob/16.x/UPGRADE.md). This is an issue for us, as we commonly apply forever coupons to subscriptions based on charitable or government status. It seems we aren't the only ones, because the [deprecation was reversed in 'clover'](https://docs.stripe.com/changelog/clover/2026-01-28/reintroduce-forever-amount-off-coupon). But until Cashier supports 'clover' or 'dahlia' we are stuck with 15.x.

Many thanks for your consideration.

sgilberg:
@crynobone @taylorotwell Pretty please can we have a 15.x release? See comment above.

lphilps:
We're having the same problem.

We'd like to upgrade to Laravel 13, while staying on Cashier v15 for the time being. The code that enables that has been checked into this repo, all that's missing is a 15.x release? Or so it seems?

PR #1846, opened 2026-06-03
Handle duplicate subscription webhooks

### LINKED ISSUE #1759: Duplicate subscriptions and no try/catch for unique Stripe ID

ibobgunardi:
Fixes #1759.

This makes the `customer.subscription.created` webhook tolerate the known race where the app's subscription creation flow and Stripe's webhook both try to persist the same subscription or subscription item.

The webhook now updates an existing local row when it is present, creates it when it is absent, and retries by reloading/updating the row only when the database error is identifiable as a unique constraint violation. Other query failures are still re-thrown.

I also added an offline regression test that simulates concurrent inserts for both the subscription and its item without requiring Stripe API credentials.

Tests run:

- `php -l src/Http/Controllers/WebhookController.php`
- `php -l tests/Feature/WebhookConcurrencyTest.php`
- `php vendor/bin/phpunit tests/Unit/WebhookControllerTest.php tests/Feature/WebhookConcurrencyTest.php`
- `php vendor/bin/phpstan analyse src/Http/Controllers/WebhookController.php tests/Feature/WebhookConcurrencyTest.php --memory-limit=1G`

### DIFF
--- a/src/Http/Controllers/WebhookController.php
+++ b/src/Http/Controllers/WebhookController.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Cashier\Http\Controllers;
 
+use Illuminate\Database\QueryException;
 use Illuminate\Http\Request;
 use Illuminate\Notifications\Notifiable;
 use Illuminate\Routing\Controller;
@@ -80,7 +81,7 @@ class WebhookController extends Controller
                 $firstItem = $data['items']['data'][0];
                 $isSinglePrice = count($data['items']['data']) === 1;
 
-                $subscription = $user->subscriptions()->updateOrCreate([
+                $subscription = $this->updateOrCreateSubscription($user, [
                     'stripe_id' => $data['id'],
                 ], [
                     'type' => $data['metadata']['type'] ?? $data['metadata']['name'] ?? $this->newSubscriptionType($payload),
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #1847, opened 2026-06-03
Resolve soft deleted subscription owners

### LINKED ISSUE #1800: Call to a member function stripe() on null (soft deletes)
### Cashier Stripe Version

16.0.4

### Laravel Version

latest

### PHP Version

8.4

### Database Driver & Version

_No response_

### Description

When receiving a webhook for a soft deleted billable model you get the error:
`Call to a member function stripe() on null`

What going good is the fact that `Cashier::findBillable()` applies `withTrashed()` to find the soft deleted model inside of the `WebhookController`.

Where it's going wrong is in the `asStripeSubscriptionItem()` method of `SubscriptionItem` looking back to the owner like so:
```php
public function asStripeSubscriptionItem(array $expand = [])
{
    return $this->subscription->owner->stripe()->subscriptionItems->retrieve(
        $this->stripe_id, ['expand' => $expand]
    );
}
```
This method is called from `$subscription->currentPeriodEnd();` which does `$item->currentPeriodEnd();`, which calls `$this->asStripeSubscriptionItem()`.

Here it loads the subscription, but it can't find the owner because it's soft deleted and there is no `withTrashed()` applied on the `subscription->owner` relation.

I think a solution would be to chaperone the subscription + owner to the `SubscriptionItem`, or add `withTrashed()` logic to the `subscription->owner` relation.

### Steps To Reproduce

1. Grab a starter kit
2. Add soft deletes to a user
3. Start a subscription
4. Delete the user
5. Cancel the subscription

ibobgunardi:
Fixes #1800.

Cashier already uses `withTrashed()` when locating a billable by Stripe ID for webhook handling, but the subscription's `owner()` relation always used a plain `belongsTo()` relation. That means a subscription item can resolve its subscription but fail to resolve the soft-deleted billable owner when paths such as `currentPeriodEnd()` call back into `asStripeSubscriptionItem()`.

This updates `Subscription::owner()` to include trashed owners only when the configured Cashier customer model uses `SoftDeletes`, preserving the existing relation for non-soft-deleting billables.

Tests run:

- `php -l src/Subscription.php`
- `php -l tests/Feature/SubscriptionOwnerTest.php`
- `php vendor/bin/phpunit tests/Unit/SubscriptionTest.php tests/Feature/SubscriptionOwnerTest.php`
- `php vendor/bin/phpstan analyse src/Subscription.php tests/Feature/SubscriptionOwnerTest.php --memory-limit=1G`

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -11,6 +11,7 @@ use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\HasMany;
+use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Collection;
 use InvalidArgumentException;
 use Laravel\Cashier\Concerns\AllowsCoupons;
@@ -93,7 +94,11 @@ class Subscription extends Model
     {
         $model = Cashier::$customerModel;
 
-        return $this->belongsTo($model, (new $model)->getForeignKey());
+        $relation = $this->belongsTo($model, (new $model)->getForeignKey());
+
+        return in_array(SoftDeletes::class, class_uses_recursive($model))
+            ? $relation->withTrashed()
+            : $relation;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #1850, opened 2026-06-03
Exclude canceled subscriptions from active status

### LINKED ISSUE #1791: Active and Canceled both return true

ibobgunardi:
Fixes #1791.

When Stripe sends a subscription update with status `canceled` but without cancel timestamps, Cashier stores `stripe_status = canceled` with `ends_at = null`. Since `active()` only checked `ends_at` and other inactive statuses, those subscriptions were still treated as active and valid.

This keeps the existing `canceled()` and grace-period semantics based on `ends_at`, but excludes Stripe's canceled status from `active()` and the active query scope.

Tests:
- `C:\laragon\bin\php\php-8.3.26-Win32-vs16-x64\php.exe vendor\bin\phpunit tests\Unit\SubscriptionTest.php tests\Feature\SubscriptionStatusTest.php`
- `C:\laragon\bin\php\php-8.3.26-Win32-vs16-x64\php.exe vendor\bin\phpstan analyse src\Subscription.php tests\Unit\SubscriptionTest.php tests\Feature\SubscriptionStatusTest.php --memory-limit=1G`
- `git diff --cached --check`

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -230,6 +230,7 @@ class Subscription extends Model
     {
         return ! $this->ended() &&
             (! Cashier::$deactivateIncomplete || $this->stripe_status !== StripeSubscription::STATUS_INCOMPLETE) &&
+            $this->stripe_status !== StripeSubscription::STATUS_CANCELED &&
             $this->stripe_status !== StripeSubscription::STATUS_INCOMPLETE_EXPIRED &&
             (! Cashier::$deactivatePastDue || $this->stripe_status !== StripeSubscription::STATUS_PAST_DUE) &&
             $this->stripe_status !== StripeSubscription::STATUS_UNPAID;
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #1860, opened 2026-07-07
Supports Stripe API codename `clover` and `dahlia` 

### LINKED ISSUE #1812: Stripe doesn't let you pick basil anymore
### Cashier Stripe Version

16.1.0

### Laravel Version

12.20.0

### PHP Version

8.2

### Database Driver & Version

_No response_

### Description

When setting up a new Stripe project and then configuring webhooks, they no longer allow you to pick basil.

Image

Would be good to go down 1 of 3 routes?

- Provide some documentation on how to force Stripe to use Basil
- Provide some documentation on how to make Cashier support Clover
- Release a Cashier update which supports Clover?

Thanks!

### Steps To Reproduce

- Install Cashier 16 on Laravel 12
- Go to stripe, set up a new account/project and set up the webhooks
- Try set API version to Basil

### LINKED ISSUE #1814: Deprecations regarding `curl_close` in Stripe PHP for PHP 8.5
### Cashier Stripe Version

16.1

### Laravel Version

12.0

### PHP Version

8.5

### Database Driver & Version

_No response_

### Description

Stripe PHP v17.3.0 uses deprecated PHP 8.5 `curl_close` function and it shows in logs and in tests on deprecation log levels.

```php
PHP Deprecated:  Function curl_close() is deprecated since 8.5, as it has no effect since PHP 8.0 in /var/www/html/vendor/stripe/stripe-php/lib/HttpClient/CurlClient.php on line 761
PHP Stack trace:
PHP   1. Stripe\HttpClient\CurlClient->__destruct() /var/www/html/vendor/stripe/stripe-php/lib/HttpClient/CurlClient.php:0
PHP   2. Stripe\HttpClient\CurlClient->closeCurlHandle() /var/www/html/vendor/stripe/stripe-php/lib/HttpClient/CurlClient.php:80
```

This issue if fixed in [stripe/stripe-php v19.1.0](https://github.com/stripe/stripe-php/releases/tag/v19.1.0) (https://github.com/stripe/stripe-php/pull/1973 Remove deprecated usage of curl_close() for PHP versions > 8.0).

How safe is it to upgrade to latest Stripe package? Is there any refactoring path or key compatibility points that need to be checked and addressed? I see that Stripe SDK v19.0.0 has changed API version to `2025-11-17.clover`.

### Steps To Reproduce

```php
$user = $request->user();

// Ensure user is a Stripe customer
if (!$user->hasStripeId()) {
    $user->createAsStripeCustomer();
}
```

crynobone:
fixes #1814
fixes #1812

### DIFF
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -99,7 +99,7 @@ class Checkout implements Arrayable, Jsonable, JsonSerializable, Responsable
             $data['cancel_url'] = $sessionOptions['cancel_url'] ?? route('home').'?checkout=cancelled';
         }
 
-        if (isset($data['ui_mode']) && StripeApiVersion::CURRENT_MAJOR === 'dahlia') {
+        if (isset($data['ui_mode']) && in_array(StripeApiVersion::CURRENT_MAJOR, ['dahlia', 'clover'])) {
             $data['ui_mode'] = match ($data['ui_mode']) {
                 'embedded' => 'embedded_page',
                 'custom' => 'elements',
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -8,8 +8,8 @@ use Illuminate\Contracts\Support\Responsable;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Support\Facades\Redirect;
 use JsonSerializable;
+use Laravel\Cashier\Enums\StripeApiVersions;
 use Stripe\Checkout\Session;
-use Stripe\Util\ApiVersion as StripeApiVersion;
 
 class Checkout implements Arrayable, Jsonable, JsonSerializable, Responsable
 {
@@ -99,12 +99,8 @@ class Checkout implements Arrayable, Jsonable, JsonSerializable, Responsable
             $data['cancel_url'] = $sessionOptions['cancel_url'] ?? route('home').'?checkout=cancelled';
         }
 
-        if (isset($data['ui_mode']) && in_array(StripeApiVersion::CURRENT_MAJOR, ['dahlia'])) {
-            $data['ui_mode'] = match ($data['ui_mode']) {
-                'embedded' => 'embedded_page',
-                'custom' => 'elements',
-                default => $data['ui_mode'],
-            };
+        if (isset($data['ui_mode'])) {
+            $data['ui_mode'] = StripeApiVersions::current()->transformUIMode($data['ui_mode']);
         }
 
         $session = $stripe->checkout->sessions->create($data);
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -99,7 +99,7 @@ class Checkout implements Arrayable, Jsonable, JsonSerializable, Responsable
             $data['cancel_url'] = $sessionOptions['cancel_url'] ?? route('home').'?checkout=cancelled';
         }
 
-        if (StripeApiVersion::CURRENT_MAJOR === 'dahlia') {
+        if (isset($data['ui_mode']) && StripeApiVersion::CURRENT_MAJOR === 'dahlia') {
             $data['ui_mode'] = match ($data['ui_mode']) {
                 'embedded' => 'embedded_page',
                 'custom' => 'elements',
--- a/src/Checkout.php
+++ b/src/Checkout.php
@@ -100,7 +100,7 @@ class Checkout implements Arrayable, Jsonable, JsonSerializable, Responsable
         }
 
         if (isset($data['ui_mode'])) {
-            $data['ui_mode'] = StripeApiVersions::current()->transformUIMode($data['ui_mode']);
+            $data['ui_mode'] = StripeApiVersions::current()->transformUiMode($data['ui_mode']);
         }
 
         $session = $stripe->checkout->sessions->create($data);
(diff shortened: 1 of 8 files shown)


src/Checkout.php
```diff
@@ -98,6 +99,14 @@ public static function create($owner, array $sessionOptions = [], array $custome
             $data['cancel_url'] = $sessionOptions['cancel_url'] ?? route('home').'?checkout=cancelled';
         }
 
+        if (isset($data['ui_mode']) && StripeApiVersion::CURRENT_MAJOR === 'dahlia') {
```
crynobone:
```suggestion
        if (isset($data['ui_mode']) && in_array(StripeApiVersion::CURRENT_MAJOR, ['dahlia', 'clover'])) {
```

src/Discount.php
```diff
@@ -30,7 +34,16 @@ public function __construct(protected StripeDiscount $discount)
      */
     public function coupon(): Coupon
     {
-        return new Coupon($this->discount->coupon);
+        $coupon = match (StripeApiVersion::CURRENT_MAJOR) {
+            'basil' => $this->discount->coupon,
+            default => $this->discount->source->coupon,
```
WendellAdriel: On clover/dahlia, stripe-php documents `source.coupon` (and `promotion.coupon`) as `null|Coupon|string`, so when the value is `null` this falls through the `instanceof` check and calls `coupons->retrieve(null)`, which blows up with an opaque SDK error, not related to the actual cause. Let's guard for `null` first and throw a clear Cashier exception naming the discount. Same applies to `src/PromotionCode.php:37`.

src/SubscriptionBuilder.php
```diff
@@ -396,6 +397,7 @@ public function checkout(array $sessionOptions = [], array $customerOptions = []
             'line_items' => Collection::make($this->items)->values()->all(),
             'mode' => 'subscription',
             'subscription_data' => array_filter([
+                'billing_mode' => $this->getBillingModeForPayload(),
```
WendellAdriel: `checkout()` merges this payload with user options via `array_merge_recursive()` on line 413, so a user passing `'subscription_data' => ['billing_mode' => ['type' => 'flexible']]` ends up with the corrupted payload `['type' => ['classic', 'flexible']]`. There's no working way to opt into flexible billing mode through `checkout()`, while `create()` allows the override via plain `array_merge()`. Let's only apply the `classic` default when the user didn't provide a `billing_mode`, or add a `billingMode()` builder method as the explicit opt-in.

src/SubscriptionBuilder.php
```diff
@@ -521,6 +524,17 @@ protected function getPriceTaxRatesForPayload(string|array $price): ?array
         return null;
     }
 
+    /**
+     * Get the billing mode for the Stripe payload.
+     */
+    protected function getBillingModeForPayload(): ?array
```
WendellAdriel: This default is only applied to builder-created subscriptions, so a subscription-mode session created through the generic path (`$user->checkout($price, ['mode' => 'subscription'])` or `Checkout::guest()`) sends no `billing_mode` and will get Stripe's new default on dahlia, diverging from `newSubscription()->checkout()`. Let's apply the same default in `Checkout::create()` when `mode === Session::MODE_SUBSCRIPTION`.

src/Checkout.php
```diff
@@ -98,6 +99,14 @@ public static function create($owner, array $sessionOptions = [], array $custome
             $data['cancel_url'] = $sessionOptions['cancel_url'] ?? route('home').'?checkout=cancelled';
         }
 
+        if (isset($data['ui_mode']) && in_array(StripeApiVersion::CURRENT_MAJOR, ['dahlia'])) {
```
WendellAdriel: This version branch is inverted compared to the other three sites in this PR: it special-cases `'dahlia'` while `Discount`, `PromotionCode`, and `SubscriptionBuilder` match on the old codenames and treat `default` as the newest behavior.

When a new version support lands in `composer.json`, this mapping silently stops applying while the others keep working. Let's rewrite it as a `match` with `'basil', 'clover' =>` keeping the old names and `default =>` mapping.

Something to consider: centralizing the codename check in one internal helper so the magic strings don't spread across four files. IDK if introducing an ENUM or something like this would cause a breaking change, but it would be nice.

WendellAdriel (changes requested):

src/SubscriptionBuilder.php
crynobone: AFAIK Checkout Session API doesn't support `billing_mode`: https://docs.stripe.com/api/checkout/sessions/create

I only see such suggestion from AI response, and I don't think it's accurate.

src/Discount.php
crynobone: Added `Laravel\Cashier\Enums\StripeApiVersions::couponFromDiscount()` and throw `InvalidArgumentException`.

Also added `couponFromPromotionCode()` method and exception for `Laravel\Cashier\PromotionCode`

src/Checkout.php
crynobone: Added `Laravel\Cashier\Enums\StripeApiVersions` enum

src/SubscriptionBuilder.php
WendellAdriel: My bad, sorry!

WendellAdriel (approved):
LGTM :muscle:

larskoole:
Any updates on this?

sts-ryan-holton:
Does this now mean I can upgrade and change to the newest stripe version? I was previously on [16.6.0](https://github.com/laravel/cashier-stripe/releases/tag/v16.6.0) and was unable to pick Basil?

PR #1868, opened 2026-08-26
Allow forever amount-off coupons

### LINKED ISSUE #1867: Client-side forever amount_off coupon guard rejects coupons Stripe accepts again since 2026-01-28.clover
### Cashier Stripe Version

16.7.0

### Laravel Version

12.67.0

### PHP Version

8.4.23

### Database Driver & Version

MySQL 8.0 (not database related)

### Description

Cashier 16 introduced client-side validation that throws `Laravel\Cashier\Exceptions\InvalidCoupon` for any `amount_off` coupon with `duration: forever`, mirroring the restriction Stripe introduced in [2025-03-31.basil](https://docs.stripe.com/changelog/basil/2025-03-31/restrict-coupon-duration).

Stripe has since **reversed that restriction** in [2026-01-28.clover](https://docs.stripe.com/changelog/clover/2026-01-28/reintroduce-forever-amount-off-coupon), and the changelog states the reversal applies *"across all API versions"*: amount-off coupons with a forever duration can be created and applied to subscriptions, invoices, checkout sessions etc. again.

Cashier's guard does not check the API version, so it still rejects these coupons client-side even though the Stripe API accepts them. The guard lives in four places:

- `SubscriptionBuilder::validateCouponForSubscriptionApplication()`
- `Subscription::applyCoupon()` (via its protected validation method)
- `Concerns\ManagesCustomer::validateCouponForCustomerApplication()`
- `Concerns\AllowsCoupons::validateCouponForCheckout()`

Each retrieves the coupon and throws when `Coupon::isForeverAmountOff()` is true — the forever check is the only validation performed there, so removing the guard restores parity with both Stripe and Cashier 15 behaviour (Stripe validates coupons server-side on apply).

This matters in practice for accounts with long-lived `amount_off` + `forever` coupons (e.g. grandfathered pricing): we have dozens of them in active use and currently have to override all the protected validation methods as no-ops to upgrade to Cashier 16.

**Suggested fix:** remove the client-side forever/amount_off guards (or gate them on the pinned Stripe API version) and let the Stripe API be the source of truth. Happy to send in a PR if that helps.

### Steps To Reproduce

1. Create an `amount_off` coupon with `duration: forever` in Stripe (allowed again on every API version since the 2026-01-28.clover reversal), e.g. via the dashboard or `stripe coupons create --amount-off 500 --currency eur --duration forever`.
2. Apply it through Cashier, e.g. `$user->newSubscription('default', 'price_x')->withCoupon('COUPON_ID')->create($paymentMethod);` or `$subscription->applyCoupon('COUPON_ID');`
3. Cashier throws `Laravel\Cashier\Exceptions\InvalidCoupon` before the subscription/discount call reaches Stripe, while the same operation performed directly against the Stripe API (`stripe subscriptions update sub_x -d "discounts[0][coupon]=COUPON_ID"`) succeeds.

UmairRathore:
Fixes #1867

Stripe reintroduced support for `amount_off` coupons with a `forever` duration in the 2026-01-28 Clover update, with the change applying across API versions. Cashier still performs the client-side validation introduced for the earlier Basil restriction, so these coupons are rejected before the subscription, customer, or Checkout request reaches Stripe.

This removes those obsolete preflight validations and lets Stripe validate coupon compatibility when the coupon is applied. The existing Stripe discount payloads are unchanged, so this only removes Cashier's stale rejection and does not alter how discounts are sent to Stripe.

This benefits applications using long-lived or grandfathered `amount_off` coupons by allowing them to upgrade to Cashier 16 without overriding the protected validation methods.

Regression coverage has been added for:

- creating a subscription with a forever amount-off coupon;
- applying one through the billable customer;
- applying one directly to an existing subscription;
- using one in Stripe Checkout.

The upgrade guidance has also been updated to reflect the restored Stripe behavior.

Locally, the full suite passes with **200 tests, 116 assertions, and 127 skipped tests**. The four new regression tests are Stripe-backed and are skipped because `STRIPE_SECRET` is not configured in my environment. I don't have access to Stripe test credentials here, so I followed Cashier's existing Stripe feature-test pattern and verified that the new tests are discovered correctly; they are ready to execute in an environment where the repository's Stripe test secret is available.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -66,9 +66,9 @@ The new `taxBehavior` method supports Basil's enhanced tax modeling (e.g., inclu
 
 ### Coupon Ending Dates
 
-Stripe's Basil API no longer supports setting discount coupons without an end date, and any application using a coupon without an end date will trigger `Laravel\Cashier\Exceptions\InvalidCoupon` exception after updating to Cashier 16.
+Stripe's Basil API originally restricted `amount_off` coupons with a `forever` duration. Stripe later reversed this restriction, and Cashier no longer rejects these coupons client-side.
 
-For existing infinite coupons, recreate them in Stripe with a finite duration before upgrading, or handle the exception by falling back to time-bound alternatives in code.
+Coupon validity is now determined by Stripe when the coupon is applied.
 
 ### Apply a Coupon by Subscription Type
 
@@ -100,7 +100,7 @@ $billable->applyPromotionCode('promotion_code_id', 'default');
 
 ### Coupon Validation in Checkouts
 
-New traits like `AllowsCoupons` introduce stricter validation for coupons in checkout sessions (e.g., ensuring compatibility with Basil). If you customize checkouts, test for rejection of invalid promotions.
+Coupon compatibility is validated by Stripe when creating Checkout sessions.
 
 ### Payment Failure Handling
 
--- a/src/Concerns/AllowsCoupons.php
+++ b/src/Concerns/AllowsCoupons.php
@@ -2,9 +2,6 @@
 
 namespace Laravel\Cashier\Concerns;
 
-use Laravel\Cashier\Coupon;
-use Laravel\Cashier\Exceptions\InvalidCoupon;
-
 trait AllowsCoupons
 {
     use InteractsWithStripe;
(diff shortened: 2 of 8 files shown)


taylorotwell:

UmairRathore:
Thanks for the feedback. I did use AI tooling during the investigation, but I understand the concern. I reviewed the affected Cashier paths and existing tests myself, but I wasn't able to execute the Stripe-backed regression cases locally, which means I shouldn't have submitted the change in that state. I'll make sure future Laravel contributions are reproduced and fully verified by me before opening a PR.

PR #1870, opened 2026-08-27
Fix subscription type after early webhook creation

### LINKED ISSUE #1773: Incorrect subscription type due to race condition
### Cashier Stripe Version

15.7.1

### Laravel Version

12.3.1

### PHP Version

8.2.12

### Database Driver & Version

_No response_

### Description

When creating a subscription with a non default type, there's a possibility it's still being saved as type 'default'.
This happens when the webhook customer.subscriptions.created is processed before the actual subscription creating function has finished.
More specifically: When executing auth()->user()->newSubscription('test')->create(), if the webhook is received and processed between lines 257 and 263 of SubscriptionBuilder.

Line 257-261:
`$stripeSubscription = $this->owner->stripe()->subscriptions->create(array_merge(
            ['customer' => $stripeCustomer->id],
            $this->buildPayload(),
            $subscriptionOptions
        ));`

Will create the subscription in Stripe

Line 263: 
`$subscription = $this->createSubscription($stripeSubscription);`

Will create the subscription in the database (if it doesn't exist yet):
`protected function createSubscription(StripeSubscription $stripeSubscription)
    {
        if ($subscription = $this->owner->subscriptions()->where('stripe_id', $stripeSubscription->id)->first()) {
            return $subscription;
        }

    ...
`

It could be circumvented by adding metadata 'type' or 'name' in Stripe, but imho this should not be required.

### Steps To Reproduce

- Create a user
- $user()->newSubscription('test')->create()

If you simulate a negative race condition by delaying line 263 in SubscriptionBuilder (by putting a breakpoint in local dev, or inserting a sleep(5), or ...), the webhook will be processed first and the subscription will be created with type 'default'

Atroci:
Fixes #1773.

When `customer.subscription.created` is handled before `SubscriptionBuilder` finishes, the webhook creates the local row with the fallback `default` type. The builder then finds that `stripe_id` and previously returned it unchanged.

This PR reconciles an existing row with the type requested by the builder. It only writes when the type differs, so repeated builder execution is a no-op.

The regression test covers:

- application write before webhook;
- webhook before application write;
- repeated webhook delivery;
- repeated builder persistence;
- one local row per Stripe subscription.

This deliberately does not change Stripe metadata behavior. The metadata approach from #1835 offers stronger recovery if the application process dies before local reconciliation, but it requires a separate decision about reserved keys and merge precedence. The broader simultaneous-insert race in #1759 also remains out of scope.

Local verification on PHP 8.4:

- `phpunit tests/Feature/SubscriptionCreationOrderTest.php`: 1 test, 8 assertions;
- unit suite plus regression: 64 tests, 108 assertions;
- targeted PHPStan: no errors.

### DIFF
--- a/src/SubscriptionBuilder.php
+++ b/src/SubscriptionBuilder.php
@@ -324,6 +324,10 @@ class SubscriptionBuilder
     protected function createSubscription(StripeSubscription $stripeSubscription)
     {
         if ($subscription = $this->owner->subscriptions()->where('stripe_id', $stripeSubscription->id)->first()) {
+            if ($subscription->type !== $this->type) {
+                $subscription->update(['type' => $this->type]);
+            }
+
             return $subscription;
         }
 
--- a/tests/Feature/SubscriptionCreationOrderTest.php
+++ b/tests/Feature/SubscriptionCreationOrderTest.php
@@ -1,0 +1,114 @@
+<?php
+
+namespace Laravel\Cashier\Tests\Feature;
+
+use App\Models\User;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Laravel\Cashier\Subscription;
+use Laravel\Cashier\SubscriptionBuilder;
+use Laravel\Cashier\Tests\TestCase;
+use Orchestra\Testbench\Attributes\WithMigration;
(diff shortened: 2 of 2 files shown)


taylorotwell:

PR #1871, opened 2026-08-30
[Issue][1862] - Upcoming invoices, check if the subscription is ended. 

### LINKED ISSUE #1862: `Subscription::upcomingInvoice` returns null on the grace period regardless of remaining billing cycle
### Cashier Stripe Version

16.5.3

### Laravel Version

13.16.1

### PHP Version

8.3.32

### Database Driver & Version

_No response_

## Problem
`Subscription::upcomingInvoice` guards on `canceled()` instead of `ended()`, which makes unable to get upcoming invoices for subscriptions on the grace period (`ends_at` is in the future)

https://github.com/laravel/cashier-stripe/blob/5dcb3b81e9b75e9770b73e7614717f6ce1d1e127/src/Subscription.php#L1289-L1299

## Why does it matter?

When using subscription schedules with `end_behavior=cancel`, in the last  Stripe marks the subscription to be canceled at period end by setting `ends_at` to future. This causes `canceled()` to be true and `upcomingInvoice()` to be null, even when there are few upcoming invoices. 

reference: https://docs.stripe.com/billing/subscriptions/subscription-schedules?locale=ja-JP#completing-schedule

### Steps To Reproduce

I am not sure all the code works but should give some sense
 ```php
// add some price / product

$product = Cashier::stripe()->products->create([
    'name' => 'unit-test-product'
]);

Cashier::stripe()->prices->create([
      'currency' => 'USD',
      'product' => $product,
      'unit_amount' => 10,
      'recurring' => ['interval' => 'month'],
      'nickname' => 'default_price',
]);

// create some user
$user = User::factory()->create();

$testClock = Cashier::stripe()->testHelpers->testClocks->create([
    'frozen_time' => now()->timestamp,
    'name' => 'feature-test-clock',
]);

$user->createAsStripeCustomer([
    'test_clock' => $testClock->id,
    'payment_method' => 'pm_card_visa',
    'invoice_settings' => ['default_payment_method' => 'pm_card_visa'],
]);

// add new subscription
$user->newSubscription('default','default_price')->add();

// create a simple schedule
$schedule = Cashier::stripe()->subscriptionSchedules->create([
            'from_subscription' => $subscriptionId,
        ]);

// add some phase data that ends in future with more than two upcoming billing cycle
$phases = [
    [
       'items' => [
          [
              'price' => $subscription->stripe_price,
              'quantity' => 1,
          ],
      ],
      'proration_behavior' => 'none',
      'start_date' => now()->timestamp,
      'end_date' => now()->addMonths(3)->timestamp,
];

// update schedule
Cashier::stripe()->subscriptionSchedules->update($schedule->id, [
            // 全てのフェーズが終わるとサブスクリプションは終了する
            'end_behavior' => 'cancel',
            'phases' => $phases,
            'proration_behavior' => 'none',
        ]);

// wait for one billing cycle (where stripe updates ends_at after invoice creation)
$testClock->advance(['frozen_time' => now()->addMonths(1)->addDay()];

// wait for subscription updated event ? 
// check upcoming invoice
$subscription->refresh()->upcomingInvoice; // should not be null
```

Liam-Broome:
Upcoming invoices should be guarded using `ended()` not `canceled()`.

In the scenario where someone has cancelled their subscription but they potentially have 3 months left this guard would have wrongfully skipped. 

So, instead we check if the subscription is ended. 

This has been tested against a feature test, a regression test was also put in place to check before and after to ensure the issue was correctly caught.

Closes: [1862](https://github.com/laravel/cashier-stripe/issues/1862)

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -1321,7 +1321,7 @@ class Subscription extends Model
      */
     public function upcomingInvoice(array $options = []): ?Invoice
     {
-        if ($this->canceled()) {
+        if ($this->ended()) {
             return null;
         }
 
--- a/tests/Feature/SubscriptionsTest.php
+++ b/tests/Feature/SubscriptionsTest.php
@@ -976,6 +976,18 @@ class SubscriptionsTest extends FeatureTestCase
         $this->assertSame(2000, $invoice->total);
     }
 
+    public function test_upcoming_invoice_is_not_null_during_grace_period()
+    {
+        $user = $this->createCustomer('upcoming_invoice_grace_period');
+
+        $subscription = $user->newSubscription('main', static::$priceId)->create('pm_card_visa');
+
+        $subscription->ends_at = now()->addMonths(3);
+        $subscription->save();
+
+        $this->assertNotNull($subscription->upcomingInvoice());
+    }
+
     public function test_updating_single_price_subscription_quantity_updates_the_quantity_of_the_subscription_item()
     {
         $user = $this->createCustomer('invoice_subscription_directly');


taylorotwell:

PR #1873, opened 2026-09-05
[16.x] Give subscription items their parent subscription

imanimen:
Subscriptions always eager load their items via `$with`, but Eloquent never sets the inverse relation, so every `SubscriptionItem` method that reaches through `$this->subscription->owner` re-queries a subscription that is already in memory.

Under `preventLazyLoading()` that surfaces as a
`LazyLoadingViolationException` when iterating subscriptions, and without it as a silent N+1: because the refetched subscription eager loads its own items in turn, each item costs two extra queries.

Handing each item the parent that just loaded it removes those queries. The items are also hidden from the item's serialized form, since the resulting circular reference would otherwise recurse without bound on Laravel 10, which has no `PreventsCircularRecursion` guard.

### DIFF
--- a/src/Subscription.php
+++ b/src/Subscription.php
@@ -107,6 +107,26 @@ class Subscription extends Model
     }
 
     /**
+     * Set the given relationship on the model.
+     *
+     * @param  string  $relation
+     * @param  mixed  $value
+     * @return $this
+     */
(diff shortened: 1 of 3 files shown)


taylorotwell:

